PR #10, opened 2024-05-02
Incidents table

### LINKED ISSUE #1: Recreate incidents-list with HTMx

podliashanyk:
Closes #1. No styling. 

The solution is heavily inspired by the [partial rendering example by django-htmx](https://github.com/adamchainz/django-htmx/blob/main/example/example/templates/partial-rendering.html).

#### NB!
Implementation relies on `django-htmx`, so activation of HtmxMiddleware on [argus-server](https://github.com/Uninett/Argus) is a **MUST**. See docs https://github.com/Uninett/argus-htmx-frontend?tab=readme-ov-file#with-extra_apps


elfjes:
wrt to columns. I could imagine something akin to the following python code. Might be more maintainable than doing everything in the row template?
```python

from argus_htmx import incident_columns, default_columns

# specifiy which columns to show and which template goes along with it
incident_columns.set(
    [
        # create a column and point to a template
        {"name": "id", "template": "htmx/incidents/_incident_pk.html"},

        # reuse a default column
        default_columns["start_time"],

        # more custom columsn
        {"name": "Site a", "template": "geant/incidents/_incident_site_a.html"},
        {
            "name": "SD Ack",
            "template": "htmx/incidents/_incident_ack.html",
            "context": {
                # we can extend the render context here perhaps?
                "ack": "sd"
            },
        },
        {
            "name": "NOC Ack",
            "template": "htmx/incidents/_incident_ack.html",
            "context": {
                # we can extend the render context here perhaps?
                "ack": "noc"
            },
        },
    ]
)
```

hmpf:
As I've been saying when it comes to customizable columns, we can crank along with just replacing the `_incidents_row.html` for now until we settle on how to do htmx and how to do updates of rows.

src/argus_htmx/templates/htmx/base.html
```diff
@@ -1,5 +1,3 @@
-{% extends "base.html" %}
```
hmpf, line 1: I don't like losing `base.html` like this.

src/argus_htmx/templates/htmx/incidents/_table.html
```diff
@@ -0,0 +1,121 @@
...
+            <th>source</th>
+            <th>description</th>
+            <th>ack</th>
+            <th>tag</th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for incident in page.object_list %}
+            <tr>
+                <td> X</td>
```
hmpf, line 19: `<td> X</td>` should also be a block.

src/argus_htmx/templates/htmx/incidents/_table.html
```diff
@@ -0,0 +1,121 @@
...
+            <th>severity</th>
+            <th>source</th>
+            <th>description</th>
+            <th>ack</th>
+            <th>tag</th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for incident in page.object_list %}
+            <tr>
```
hmpf, line 18: The whole row should be in a block.

src/argus_htmx/templates/htmx/incidents/_table.html
hmpf: Open question if the tag itself should be in the block or not, if yes it would be easy to add classes/ids to the tag. It would also make it easier to replace the table with  div-in-div soo,, pros/cons...

hmpf: Anyway, each cell should have their own id at least.

src/argus_htmx/templates/htmx/incidents/incidents.html
hmpf, line 1: This file should replace `htmx/incidents/list.html` eventually.

We need to think of a naming convention seeing as we also have "detail.httml".

src/argus_htmx/urls.py
```diff
@@ -19,4 +19,5 @@
     path("incidents/", views.incidents, name="htmx_incidents"),
     path("incidents/<int:pk>/", views.incident_detail, name="htmx_incident_detail"),
     path("incidents/<int:pk>/ack/", views.incident_add_ack, name="htmx-incident-add-ack"),
+    path("incidents/table/", views.incidents_table, name="htmx_incidents_table"),
```
hmpf, line 22: To replace` path("incidents/", views.incidents, name="htmx_incidents"),` eventually.

src/argus_htmx/views.py
```diff
@@ -65,5 +71,41 @@ def incident_add_ack(request, pk: int, group: Optional[str] = None):
...
+    page_num = request.GET.get("page", "1")
+    page = Paginator(object_list=qs, per_page=10).get_page(page_num)
+
+    # The htmx magic - use a different, minimal base template for htmx
+    # requests, allowing us to skip rendering the unchanging parts of the
+    # template.
+    if request.htmx:
+        base_template = "htmx/incidents/_table.html"
+    else:
+        base_template = "htmx/incidents/_base.html"
```
hmpf, line 94: How will this work with filters? We could add just one as a placeholder. Acked, say, adding more stuff to `request.GET`.

hmpf (changes requested):
More cowb^Wblocks, + some thoughts.

src/argus_htmx/templates/htmx/base.html
```diff
@@ -31,3 +31,7 @@ <h1><span>Logo</span> Argus: {{ page_title }}</h1>
 {% endblock %}</div>
 </header>
 {% endblock header %}
+
+<main>
+    {% block main %}{% endblock %}
+</main>
```
hmpf, line 37: I suspect this doesn't do what you think it does.

src/argus_htmx/templates/htmx/incidents/_incident_table_row.html
```diff
@@ -0,0 +1,41 @@
+<tr
+        id="incident-{{ incident.pk }}-row"
+>
+    <td>
+        {% block bulk_operations %}
+            <p>X</p>
+        {% endblock bulk_operations %}
+    </td>
```
hmpf, line 8: Maybe "chooser" instead of "bulk_operations"?

src/argus_htmx/templates/htmx/incidents/_incident_table_row.html
```diff
@@ -0,0 +1,41 @@
...
+            {% include "htmx/incidents/_incident_start_time.html" %}
+        {% endblock %}
+    </td>
+    <td>
+        {% block status %}
+            {% include "htmx/incidents/_incident_status.html" %}
+        {% endblock %}
+    </td>
+    <td>{{ incident.level }}</td>
+    <td>{{ incident.source }}</td>
```
hmpf, line 25: Remember that Géant won't need this column at all.

src/argus_htmx/templates/htmx/incidents/_incident_table_row.html
```diff
@@ -0,0 +1,41 @@
...
+        {% block start_time %}
+            {% include "htmx/incidents/_incident_start_time.html" %}
+        {% endblock %}
+    </td>
+    <td>
+        {% block status %}
+            {% include "htmx/incidents/_incident_status.html" %}
+        {% endblock %}
+    </td>
+    <td>{{ incident.level }}</td>
```
hmpf, line 24: I think Géant uses colors in some other columns for this, not a separate column.

src/argus_htmx/templates/htmx/incidents/incidents.html
hmpf, line 1: Didn't we decide to rename this to `incidents_list.html`?

src/argus_htmx/templates/htmx/incidents/_table.html
```diff
@@ -0,0 +1,89 @@
...
+        <tr>
+            <th>actions</th>
+            <th>id</th>
+            <th>timestamp</th>
+            <th>status</th>
+            <th>severity</th>
+            <th>source</th>
+            <th>description</th>
+            <th>ack</th>
+            <th>tag</th>
```
hmpf, line 13: These need to be swappable. New block "incident_table_header" covering all of them or something..

hmpf (changes requested):

src/argus_htmx/urls.py
hmpf: I'm not really working on lists anymore. I think we'll merge this as is then replace lists with it.

PR #15, opened 2024-05-08
Add theming-support

hmpf:
Depends on Uninett/Argus#810

Use a context processor to set the variable "theme" in all contexts. The variable corresponds to a file in `static/color-schemes/`.

Included themes: dark, light, blue, where dark supports darkmode and light is the default.

How to use: add one of the included context processors either depending on #810 or manually.

Testing with GET: use the "argus_htmx.context_processors.theme_via_GET" context processor and suffix `?theme=SOMECOLOR` to any url.

Testing with session: use the "argus_htmx.context_processors.theme_via_session" context processor and set the correct key in your session like so: `request.session["theme"] = "SOMECOLOR"`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,6 +45,11 @@ In your local settings that star-imports from an `argus-server`_ settings file::
     ROOT_URLCONF = "urls.py"
     MIDDLEWARE += ["django_htmx.middleware.HtmxMiddleware"]
 
+In the same file, add a copy of the entirety of ``TEMPLATES``. Choose one of
+the functions in ``argus_htmx.context_processors``. In the entry for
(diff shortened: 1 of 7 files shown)


hmpf:
"blue" is here as an example, we should probably ship with something crazy like "startrek" or "reggae" or "80ies" in the final version of this app. Crazy ideas welcome!

elfjes:
I like having the ability to provide an arbitrary css file to add as a stylesheet :)
Maybe we can even have multiple? themes: `blue` and `geant`? the `blue` theme would set color variables like it does now, and the geant theme would be more specific and certain specific classes/selectors.

hmpf:
See #17 for theme-issue

PR #18, opened 2024-05-15
Incidents table poll

podliashanyk:
Closes Uninett/argus-htmx-frontend#2

### Implemented changes:
- incidents table is polled every 30 seconds. Whole contents of table body are being replaced with table rows comprised from the incident list returned by the endpoint
- added table refresh info
- standard Django pagination is used, meaning that changes to incidents count might move incidents that are currently in users view to a page that is not actively in user's view

### Further work:
- Uninett/Argus#989

### Notes:
- Granularity of the reswap is all table rows that are currently in user's view. This is pretty coarse yet cheap per now. In the future more granular swaps might be considered. Thus [multi-swap extension](https://htmx.org/extensions/multi-swap/), or [out-of-bound swaps](https://htmx.org/attributes/hx-swap-oob/) should be consider as a replacement for current swap method.
- More updates can be 'piggy-backed' simlply by adding [out-of-bound swap tags](https://htmx.org/attributes/hx-swap-oob/) to the `_incidents_table_poll.html`.
- Poll might be moved to a separate endpoint in the future. Per now its logic is simple enough and resembles normal table render.

### Extra attention/feedback/segguestions needed:
- Refresh interval is hardcoded (30 seconds). How should we configure this value? Default per Arhus installation, but overridable per user? Where to store this config?
- Time is ripe to require tests for implemented changes in PRs?
- Should we already start considering infinite scroll as an alternative table view?
- If pagination method is to be changed in the future, the HTMX polling logic (`_incidents_table_poll.html` and `<table>` attributes) wont't be affected. Meaning that this PS doesn't need to concern those changes too much, right? Or are there other considerations to keep in mind?


hmpf (approved):
Works

Issue #4, opened 2024-04-22
Make it possible to configure which columns are visible


hmpf:
Easy to swap between:

* Argus-style
* Geant-style, which needs some tags as columns, separate ack-column, separate ticket column

hmpf:
First round merged in #31.

Needed: document how to support specific geant needs

PR #25, opened 2024-06-25
Make columns customizable

podliashanyk:
For #4

Depends on https://github.com/Uninett/Argus/pull/838


src/argus_htmx/incidents/customization.py
```diff
@@ -0,0 +1,61 @@
...
+    ("links", IncidentField("links", "Actions", None))
+])
+
+# TODO remove before alpha release
+# Fields for demo/debug/dev purposes
+TEMP_FIELDS = {
+    "id": IncidentField("id", "ID", "htmx/incidents/_incident_pk.html"),
+    "tag": IncidentField("tag", "Tag", "htmx/incidents/_incident_tag.html", context={"tag": "location"}),
+    "ack": IncidentField("ack", "Ack", "htmx/incidents/_incident_ack.html")
+}
```
hmpf, line 43: These are all useful even for us. Rename to OTHER_FIELDS or EXTRA_FIELDS instead?

hmpf: Looking closer, the "tag" IncidentField needs to have customizable key.

We have use of an ack-column, geant has use of a grouped ack-column.

hmpf (commented):
This will allow for customization yes, but how does one actually customize? Another PR after this one?

If a setting we already have support for json in settings due to the app plugin machinery.

Remember we need "dynamic" fields for tags and acks. Tags seems easy, we look up a specific key in the tags-list. Acks is worse, we either show any ack OR look up an ack with a specific group. Have an example of both, in ADDITIONAL/EXTRA/OTHER_FIELDS?

Geant may need to look into the black box that is the metadata blob (though I hope they use tags). @saket-geant?

src/argus_htmx/incidents/customization.py
hmpf: The problem is the lookup needed. This lookup: `incident.acks.active().filter(group__name=GROUP)` cannot/should not be done in a template.

The tags lookup already works.

hmpf:
Json settings: `argus.site.settings.get_json_env()`, and it will need a validator. The app plugins use pydantic, see `argus.site.settings.validate_app_setting()` and `argus.site.settings._serializers.ListAppSetting`.

README.rst
```diff
@@ -92,6 +92,58 @@ Icons, pictures, etc. used by the theme MUST be in a subdirectory named the
...
+        "name": str,  # identifier
+        "label": str,  # display value
+        "cell_template": str  # path to template
+
+        "context": Optional[dict]  # context for the template
+    }
+
+Export an environment variable to configure columns that are displayed::
+
+    ARGUS_UI='{"table_fields": [{"name": "id", "label": "id2", "cell_template": "htmx/incidents/_incident_pk.html"}, ...]}'
```
hmpf, line 119: .. I do

```
export ARGUS_EXTRA_APPS=`cat extra.json`
```

in the shellscript I use to wrap manage.py. "extra.json" is primly indented for maximum human editability =)

src/argus_htmx/incidents/customization.py
```diff
@@ -11,7 +11,15 @@
...
-from typing import Optional, OrderedDict
+from typing import Optional, OrderedDict, Union
+
+
+@dataclass
+class FieldContext:
+    """Class for providing additional context to a dynamic field."""
+
+    ack: Optional[str] = None  # team or user group
+    tag: Optional[dict] = None  # incident tag name
```
hmpf, line 22: `# team (via user group)`

src/argus_htmx/incidents/customization.py
```diff
@@ -38,8 +46,18 @@ class IncidentField:
...
+EXTRA_FIELDS = {
+    "location_tag": IncidentField("location", "Location", "htmx/incidents/_incident_tag.html",
+                                  context=FieldContext(tag={"key": "location"})),
+    "problem_type_tag": IncidentField("problem_type", "Problem Type", "htmx/incidents/_incident_tag.html",
+                                      context=FieldContext(tag={"key": "problem_type"})),
+    "any_ack": IncidentField("ack", "Ack", "htmx/incidents/_incident_ack.html"),
+    "sd_ack": IncidentField("sd_ack", "SD Ack", "htmx/incidents/_incident_ack.html",
+                            context=FieldContext(ack="sd")),
+    "noc_ack": IncidentField("noc_ack", "NOC Ack", "htmx/incidents/_incident_ack.html",
+                             context=FieldContext(ack="noc")),
```
hmpf, line 60: This is hardcoding both ours and geant's ack-systems. Hardcode ours, as just "ack", show how to do geant's in doc.

src/argus_htmx/settings/ui_settings.py
```diff
@@ -0,0 +1,13 @@
...
+from argus_htmx.settings import validate_ui_setting
+
+_ui_env = get_json_env("ARGUS_UI", {}, quiet=False)
+UI = validate_ui_setting(_ui_env)
+del _ui_env
+
+# customize the displayed columns in the incident table
+TABLE_FIELDS = IncidentFields()  # initialize with default fields
+if UI and UI.table_fields:
+    TABLE_FIELDS.set_fields(UI.table_fields)  # override table fields from JSON setting
```
hmpf, line 13: Does this *replace*, *add* or *unify with* the existing settings? Mention it in a comment in the code.

"replace" is better than "override" if nothing of the default remains.

src/argus_htmx/templates/htmx/incidents/_incident_ack.html
```diff
@@ -1,3 +1,5 @@
+{% load argus_htmx %}
+
 <a href="{% url 'htmx:incident-add-ack' pk=incident.pk %}">
-	{% if incident.acked %}Acked{% else %}Unacked{% endif %}
+    {% if incident|isackedby:field.context %}Acked{% else %}Unacked{% endif %}
```
hmpf, line 4: Yep, only way to do it in django. :thumbsup:

"is_acked_by" looks better though.

src/argus_htmx/templates/htmx/incidents/_incident_tag.html
```diff
@@ -1,2 +1,2 @@
 {% load argus_htmx %}
-{{ incident|tagvalues:tag|first }}
+{{ incident|tagvalues:field.context.tag.key|first }}
```
hmpf, line 2: The reason "tagvalues" doesn't have an underscore is because it is a noun, but django template filters can't have a dash (-) in the name.

You might consider renaming "isackedby" to "ackgroup", or "tagvalues" to "has_tag_key"

src/argus_htmx/settings/ui_settings.py
hmpf: Template filters can be great fun! Much easier than making template tags.

hmpf (changes requested):
There ought to be tests but the "setting up testing for 3rd party app that allow choosing which tests to run" branch is languishing, outdated, on my work computer. We need an issue for tests for this, at minimum:

* check that the settings override/replacement works
* check that the template filters work

See also the other comments.

src/argus_htmx/incidents/customization.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_fields(self):
+        return self.fields
+
+    def set_fields(self, fields: list[IncidentField]):
+        self.fields = fields
+
+    def get_merged_context(self):
+        return {k: v for field in self.fields if field.context for k, v in
+                (field.context.__dict__.items() if isinstance(field.context, FieldContext) else field.context.items())}
```
elfjes, line 77: I'm probably missing something here, but when we have supplied multiple fields with duplicate context keys having different values (ack="noc_ack" and ack="sd_ack", does this collapse those duplicates so that only a single ack="..." context value is retained? How do we then make a distinction in the separate column templates?

src/argus_htmx/templatetags/argus_htmx.py
```diff
@@ -11,4 +11,22 @@ def tagvalues(incident, key) -> list:
...
+
+    if not context:
+        is_acked = incident.acked
+    else:
+        if context.ack is None:
+            is_acked = incident.acked
+        else:
+            is_acked = context.ack in incident.acks.active().group_names()
+
+    return is_acked
```
hmpf, line 32: Sikt just needs a template that checks if `incident.acked` is True.

Geant needs a template that checks if `incident.acked` is True and also checks if it belongs to a specific group.

This filter doesn't work well for either, it is hard to understand and over-abstracted. It would be better in this one case to either *not* have this filter and expect Geant to make their own or make a filter that returns True if this incident is acked and was acked by a user of a specific group.

src/argus_htmx/templates/htmx/incidents/_incident_ack.html
```diff
@@ -1,3 +1,5 @@
+{% load argus_htmx %}
+
 <a href="{% url 'htmx:incident-add-ack' pk=incident.pk %}">
-	{% if incident.acked %}Acked{% else %}Unacked{% endif %}
+    {% if incident|is_acked_by:field.context %}Acked{% else %}Unacked{% endif %}
```
hmpf, line 4: Not needed in the general case. See other comment.

hmpf (changes requested):

src/argus_htmx/incidents/customization.py
hmpf: This might be a case of "typing ruins everything"..

elfjes:
I've created an new PR #31 for column customization that reduces the scope and simplifies this a bit so to hopefully merge it sooner

hmpf:
Partially replaced by #31

PR #28, opened 2024-06-27
Optimize first load of /incidents/

hmpf:


src/argus_htmx/incidents/views.py
```diff
@@ -23,8 +23,9 @@
 def prefetch_incident_daughters():
     return (
         Incident.objects
-        .select_related("source")
+        .defer("metadata", "search_text")
         .prefetch_related(
```
elfjes, line 27: I'm not in favor of deferring metadata loading in the incidents listing since we will need the metadata there. I'm concerned that this might lead to n+1 queries since now we will read/query metadata in the template

elfjes (commented):
What is the purpose of all the css changes in this PR?

elfjes:
Furthermore, I'm seeing this n+1 issue in the incident table,  where the below queries is executed for every incident in the current table page

```
2024-07-01 08:34:21.124 UTC [46] LOG:  statement: SELECT "argus_incident_sourcesystemtype"."name" FROM "argus_incident_sourcesystemtype" WHERE "argus_incident_sourcesystemtype"."name" = 'correlator' LIMIT 21
2024-07-01 08:34:21.126 UTC [46] LOG:  statement: SELECT 1 AS "a" FROM "argus_incident_event" LEFT OUTER JOIN "argus_incident_acknowledgement" ON ("argus_incident_event"."id" = "argus_incident_acknowledgement"."event_id") WHERE ("argus_incident_event"."incident_id" = 20 AND (("argus_incident_acknowledgement"."event_id" IS NOT NULL AND ("argus_incident_acknowledgement"."expiration" IS NULL OR "argus_incident_acknowledgement"."expiration" > '2024-07-01T08:34:21.125522+00:00'::timestamptz)) OR ("argus_incident_event"."type" = 'ACK' AND "argus_incident_acknowledgement"."event_id" IS NULL))) LIMIT 1

``` 

we may want to prefetch some more things

hmpf:
This is totally outdated so closing.

PR #30, opened 2024-07-22
Add fonts and remove source map urls for themes

### LINKED ISSUE #29: collectstatic fails due to missing urls from themes
The current themes refer to some urls (font, source maps) that do not resolve from the theme location. ``./cmd.sh collectstatic`` tries to resolve these files and crashes. eg

```
ValueError: The file 'themes/dark.css.map' could not be found with <django.contrib.staticfiles.storage.ManifestStaticFilesStorage object at 0x7f1d9298ac10>.
```

We should either add the missing urls, or remove them from the themes...

elfjes:
* Add font files for et-book theme
* remove source map references for
  * bahunya
  * kacit
  * tacit
  * water-dark
  * water-light

fixes #29


hmpf (approved):

hmpf:
I would have preferred one commit for the font and one for the source map removal but we're in a hurry so let's go!

PR #36, opened 2024-07-24
Setup testing with tox

hmpf:
This setup will work if *not* testing anything that uses imports from argus-server. If importing anything from argus-server it becomes necessary to expand the settings-file and switch to an actual postgres as the test database.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,0 +1,15 @@
+#!/usr/bin/env python
+import os
+import sys
+
+import django
+from django.conf import settings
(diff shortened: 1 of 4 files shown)


elfjes:
What are your thoughts on using pytest?

hmpf:
Pytest as a test runner is fine but I prefer the unittest style of writing the actual tests. Especially for Django since Django's own DjangoTestCase is very nice.

Something like this runtests.py I have used in plenty of standalone django apps.

PR #37, opened 2024-07-24
Add middleware for forcing logins, w/tests

hmpf:
For this to work, change the following settings:
```
LOGIN_URL = "/accounts/login"
PUBLIC_URLS = [
    "/accounts/login",
    "/api",
]
MIDDLEWARE += ["argus_htmx.middleware.LoginRequiredMiddleware"]
```

Urls in `PUBLIC_URLS` are ignored by the middleware. Likewise, if the view has the attribute `login_required` set to "False", this middleware is skipped. If neither holds an unauthenticated user is redirected to LOGIN_URL.

(argus-server needs to mark "/" as "login_required = False" and any public urls not in the API.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "argus-server",
+    "argus-server>=1.20.0",
     "django-htmx",
     # The next is for python_version < 3.12 but we simplify code by always needing it
     "importlib_resources>=5.12",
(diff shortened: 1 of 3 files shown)


hmpf:
This PR depends on argus-server 1.20.0 or newer, which has already marked the html pages it has as public.

With the support for middleware in EXTRA_APPS-setting it is also possible to change the argus_htmx entry from

```
  {
    "app_name": "argus_htmx",
    "urls": {
      "path": "",
      "urlpatterns_module": "argus_htmx.urls"
    },
    "context_processors": [
      "argus_htmx.context_processors.theme_via_session"
    ]
  },
```

to 

```
  {
    "app_name": "argus_htmx",
    "urls": {
      "path": "",
      "urlpatterns_module": "argus_htmx.urls"
    },
    "context_processors": [
      "argus_htmx.context_processors.theme_via_session"
    ],
    "middleware": {
      "argus_htmx.middleware.LoginRequiredMiddleware": "end"
    }
  },
```

PR #38, opened 2024-07-26
Add box for filtering incidents to incident-list

hmpf:
How to swap out the box:

Set the setting "ARGUS_HTMX_FILTER_FUNCTION" to the dotted path of a module that contains the function "incident_list_filter", see argus_htmx/incidents/filter.py for signature.

You might then not need to replace the template: htmx/incidents/_incident_filterbox.html

(Replacing the template *will* be necessary to make things pretty.)

### DIFF
--- a/src/argus_htmx/incidents/filter.py
+++ b/src/argus_htmx/incidents/filter.py
@@ -1,0 +1,70 @@
+from django import forms
+
+from argus.filter import get_filter_backend
+from argus.incident.constants import INCIDENT_LEVEL_CHOICES, MAX_INCIDENT_LEVEL
+from argus.incident.models import SourceSystem
+
+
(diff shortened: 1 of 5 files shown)


elfjes:
I would suggest using the existing filtering plugin system instead of a new `ARGUS_HTMX_FILTER_FUNCTION` setting, and perhaps split the function call for getting an incident form and the filtered qs to two separate functions?

elfjes:
Also, let's maintain api compatibility with `BaseFilterBackend`/`DjangoFilterBackend`: 
```python
class BaseFilterBackend:
    def filter_queryset(self, request, queryset, view)-> QuerySet:
        ...
```

It won't harm, and it can help further integration with drf

PR #43, opened 2024-08-19
Basic customizable tailwind styles

podliashanyk:
### Relevant stack:
- https://github.com/dobicinaitis/tailwind-cli-extra
- https://daisyui.com/theme-generator/
- https://pypi.org/project/django-widget-tweaks/

### UI with daisyUI theme "forest":
Screenshot 2024-08-19 at 13 36 15

### DIFF
--- a/src/argus_htmx/tailwindtheme/tailwind.config.js
+++ b/src/argus_htmx/tailwindtheme/tailwind.config.js
@@ -1,0 +1,47 @@
+const projectPaths = [
+    '../templates/**/*.html',
+];
+
+const contentPaths = [...projectPaths];
+console.log(`tailwindcss will scan ${contentPaths}`);
+
+module.exports = {
+    content: contentPaths,
+    theme: {
+        borderWidth: {
+            DEFAULT: '2px',
+        },
+        ...{},
+    },
+    daisyui: {
+        themes: ['dark', 'light', {
+            'argus': {
+                'primary': '#006d91',
+                'primary-content': '#d1e1e9',
+                'secondary': '#f3b61f',
+                'secondary-content': '#140c00',
+                'accent': '#c84700',
+                'accent-content': '#f8dbd1',
+                'neutral': '#006d91',
+                'neutral-content': '#d1e1e9',
+                'base-100': '#edfaff',
+                'base-200': '#ced9de',
+                'base-300': '#b0babd',
+                'base-content': '#141516',
+                'info': '#0073e5',
+                'info-content': '#000512',
+                'success': '#008700',
+                'success-content': '#d3e7d1',
+                'warning': '#ee4900',
+                'warning-content': '#140200',
+                'error': '#e5545a',
+                'error-content': '#120203'
+            }
+        }],
+    },
+    plugins: [
+        require('@tailwindcss/forms'),
+        require('@tailwindcss/typography'),
+        require("daisyui"),
+    ]
+}


elfjes:
Nice work and thanks for sharing this in its early stage 🙌 

I think in order to combine css we need to figure out the following:

We need to be able to run our own `tailwindcss` combining our css (overrides) with yours. Currently you have two css files that we'd need to include: `argus_htmx/static/tailwindtheme/overrides.css` and `argus_htmx/tailwindtheme/main.css`. Because of the `@apply` directive we need to include them prior to building our tailwind.css. However, locating these files dynamically is kind of tricky (having to do with venvs, site-packages and the likes). 

It would be nice if these files could perhaps be compiled separately so that they can be checked into git as a (smal) regular css file and django-static would find them for us. You'd then need a separate tailwind build for your complete sources (which you wouldn't want to check into git because it's too big and generated dynamically). Same as us, we'd also need to build our tailwind against all our sources. But at least that would make discovery of your tailwind customizations more simple. Or, if it's not too much hassle and works properly, could you create those overrides without @apply so that they wouldn't need a build step at all? (I'm not quite sure how that would work with daisy themes though).

Another question: Do we/you really need those @apply overrides? It looks like you can just add those tailwind classes to the html-nodes that currently have that class? That would also greatly simplify the css build step

Alternatively, we could take a similar approach as with dynamically finding template dirs, and expose a django setting pointing to a list of css files to include in a tailwind css build step (and we would extend that list with our files. We can then dynamically generate the `styles.css` like we do with `tailwind.config.js`

src/argus_htmx/static/styles.css
```diff
@@ -0,0 +1,3677 @@
+/*
+! tailwindcss v3.4.10 | MIT License | https://tailwindcss.com
+*/
+
+/*
+1. Prevent padding and border from affecting element width. (https://github.com/mozdevs/cssremedy/issues/4)
+2. Allow adding a border to an element by just adding a border-width. (https://github.com/tailwindcss/tailwindcss/pull/116)
+*/
```
elfjes, line 8: You probably already know and this is most likely due to pushing as early as you did, but you probably don't want to check in a compiled tailwind css -> .gitignore

src/argus_htmx/templates/htmx_base.html
```diff
@@ -1,26 +1,29 @@
...
-  <meta charset="utf-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1">
-  {% block title %}<title>Argus Server: {{ page_title }}</title>{% endblock %}
-  {% block head %}
-  {% endblock head %}
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    {% block title %}<title>Argus Server: {{ page_title }}</title>{% endblock %}
+
+    <link rel="stylesheet" href="/static/styles.css">
```
elfjes, line 9: can we get this as a non-hardcoded css link?

src/argus_htmx/tailwindtheme/tailwind.config.js
```diff
@@ -0,0 +1,89 @@
...
+    },
+    daisyui: {
+        themes: [...daisyUIThemes], // false: only light + dark | true: all themes | array: specific themes like this ["light", "dark", "cupcake"]
+        darkTheme: "dark", // name of one of the included themes for dark mode
+        base: true, // applies background color and foreground color for root element by default
+        styled: true, // include daisyUI colors and design decisions for all components
+        utils: true, // adds responsive and modifier utility classes
+        prefix: "", // prefix for daisyUI classnames (components, modifiers and responsive class names. Not colors)
+        logs: true, // Shows info about daisyUI version and used config in the console when building your CSS
+        themeRoot: ":root", // The element that receives theme color CSS variables
```
elfjes, line 82: Just curious: how much of this is different from the defaults?

elfjes (commented):
Like I said earlier, thanks for the wip push :). In general I really like the work you've done but it needs some attention to make sure that we can integrate our own themes into daisy/tailwind (see also my earlier comment)

elfjes:
> > We need to be able to run our own `tailwindcss` combining our css (overrides) with yours. Currently you have two css files that we'd need to include: `argus_htmx/static/tailwindtheme/overrides.css` and `argus_htmx/tailwindtheme/main.css`. Because of the `@apply` directives we need to include them prior to building our tailwind.css. However, locating these files dynamically is kind of tricky (having to do with venvs, site-packages and the likes).
> > It would be nice if these files could perhaps be compiled separately so that they can be checked into git as a (smal) regular css file and django-static would find them for us. You'd then need a separate tailwind build for your complete sources (which you wouldn't want to check into git because it's too big and generated dynamically). Same as us, we'd also need to build our tailwind against all our sources. But at least that would make discovery of your tailwind customizations more simple. Or, if it's not too much hassle and works properly, could you create those overrides without `@apply` so that they wouldn't need a build step at all? (I'm not quite sure how that would work with daisy themes though).
> > Another question: Do we/you really need those `@apply` overrides? It looks like you can just add those tailwind classes to the html-nodes that currently have that class? That would also greatly simplify the css build step
> 
> The `argus_htmx/tailwindtheme/main.css` one is temporary. I used it in order to be able to experiment with tailwind styles from one file as opposed to numerous `html` files. All of the `@apply` classes in `argus_htmx/tailwindtheme/main.css` will be moved to relevant `html` files. The `argus_htmx/static/tailwindtheme/overrides.css` one I did intend to include. It can somewhat be replaced by adding a custom `<link rel="stylesheet" ...>` to the overridable `head` block in html. The latter option may lead to some unpredictable results but we could land on it for the time being nonetheless. This way we would also see if we even have a need to have a more controllable overrides. :)

alright, great 🎉 . Curious what the result will be :)

src/argus_htmx/static/styles.css
```diff
@@ -1,3677 +0,0 @@
-/*
-! tailwindcss v3.4.10 | MIT License | https://tailwindcss.com
-*/
-
-/*
```
elfjes: Don't forget to squash your commits (eg when merging), otherwise this big file remains in the git history

src/argus_htmx/tailwindtheme/styles.css
```diff
@@ -1,3 +1 @@
 @import "tw.css";
```
elfjes, line 1: Perhaps you want to inline the contents of tw.css into here, given that all it's doing is including tailwind

src/argus_htmx/templates/htmx_base.html
```diff
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% block title %}<title>Argus Server: {{ page_title }}</title>{% endblock %}
 
-    <link rel="stylesheet" href="/static/styles.css">
+    <link rel="stylesheet" href="{% static path_to_stylesheet|default:"styles.css" %}">
```
elfjes, line 9: nice 👍

.gitignore
```diff
@@ -9,6 +9,13 @@ __pycache__/
 build/
 dist/
 
+# Tailwind CSS
+src/argus_htmx/static/styles.css
+# Autogenerated via management command
+src/argus_htmx/tailwindtheme/tailwind.config.js
+# OS specific
+src/argus_htmx/tailwindtheme/tailwindcss
+
```
hmpf, line 18: If these are to be served with the finished package, they cannot be in .gitignore, because argus proper uses setuptools_scm to build a package. This will not include anything hit by .gitignore.

elfjes: yes, don't include `tailwindcss` in the package.

I think you do need to include the styles.css in the package. Otherwise you'd need to perform some post-installation work to generate it everywhere you install the package (such as in a deployed environment). Which *could* be done (for example,. extend the `tailwind_config` command), but you need to ensure that you have a `tailwindcss` for your platform installed.

Is there no possibility to have setuptools-scm make an exception for a specific file and include it anyway?

README.rst
```diff
@@ -77,21 +78,99 @@ Update
...
+Overview
+~~~~~~~~
+* Tailwind CSS: A utility-first CSS framework for rapidly building custom user interfaces.
+* daisyUI: A component library for Tailwind CSS that provides a set of ready-to-use components as well as color themes.
+
+Installation and build instructions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Recommended but open for tweaks and adaptations steps:
+
+1. Install Tailwind standalone CLI bundled with daisyUI from https://github.com/dobicinaitis/tailwind-cli-extra as ``tailwindcss`` executable. Optionally you can compile tailwind+daisyUI standalone cli bundle yourself as described here: https://github.com/tailwindlabs/tailwindcss/discussions/12294#discussioncomment-8268378.
```
hmpf, line 102: We should include this recipe in our docs. In a documentation PR.

README.rst
```diff
@@ -77,21 +78,99 @@ Update
...
+2. Generate Tailwind config file by running ``tailwind_config`` management command. By default the generated file will be based on ``src/argus_htmx/tailwindtheme/tailwind.config.template.js`` and expected values will be injected with reasonable defaults. For customization options see `Customization`_.
```
hmpf, line 103: We already supply a config-file right, or is that os-dependent too?

In the docs we should have a howto for how to upgrade tailwind++ and customize tailwind but I don't think step 2 belongs in the README. We'll move it in a docs-PR

README.rst
```diff
@@ -77,21 +78,99 @@ Update
...
+3. Build main stylesheet file using ``tailwindcss`` executable from step 1 and pointing to config file from step 2::
+
+    ./tailwindcss -c tailwind.config.js -i ./styles.css --output ../static/styles.css
+   optional ``--watch`` flag may be included if you want the ``--output`` CSS file being rebuild on changes in template files.
```
hmpf, line 107: Now *this* (especially --watch) is important.

README.rst
```diff
@@ -77,21 +78,99 @@ Update
...
+
+
+Customization
+~~~~~~~~~~~~~
+Following are a subject to customization:
+
```
hmpf, line 113: All of this section should go in a separate file. It is not daily bread and butter dev. In another PR.

hmpf (commented):
I'll test manually. Please rebase, as I think it would be wise of me to go commit by commit in this one.

hmpf (commented):
Commit by commit I would say the "Add more styles"-commit is a bit messy (it also configures tailwind) but let's leave it as is. Now for the manual test!

hmpf (approved):
The stuff just needed for dev should not be in src/. I'd say put it in tailwind/.

PR #61, opened 2024-08-22
Add logo to navbar

podliashanyk:
Adaptation of #47 after #43

### DIFF
--- a/src/argus_htmx/templates/htmx/base.html
+++ b/src/argus_htmx/templates/htmx/base.html
@@ -14,7 +14,7 @@
         <nav class="navbar">
             <div class="navbar-start">
                 <a class="link" href="{% url 'htmx:incident-list' %}">
-                    <img class="object-scale-down h-14" src="{% static "logo_white.svg" %}" alt="Argus logo">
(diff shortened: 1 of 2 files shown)


src/argus_htmx/templates/htmx/base.html
```diff
@@ -12,7 +13,9 @@
     <header class="bg-neutral text-neutral-content">
         <nav class="navbar">
             <div class="navbar-start">
-                <a class="btn btn-ghost text-xl" href="{% url 'htmx:incident-list' %}">Argus Logo</a>
+                <a class="link" href="{% url 'htmx:incident-list' %}">
+                    <img class="object-scale-down h-14" src="{% static "logo_white.svg" %}" alt="Argus logo">
```
hmpf, line 17: I think just "Argus" or "Home" as the alt text is better.

hmpf: If we write "Argus" after the logo we could hide the logo completely from screen readers.

hmpf (approved):

PR #62, opened 2024-08-22
Convert user section in navbar to an avatar with dropdown menu

### LINKED ISSUE #57: Clicking on username in the header should result in a dropdown, not expanding the header

podliashanyk:
Closes #57

New menu looks like this when expanded:

Screenshot 2024-08-22 at 15 27 26

### DIFF
--- a/src/argus_htmx/static/styles.css
+++ b/src/argus_htmx/static/styles.css
@@ -1,0 +1,4278 @@
+/*
+! tailwindcss v3.4.10 | MIT License | https://tailwindcss.com
+*/
+
+/*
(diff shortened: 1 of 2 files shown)


hmpf:
> > It does not open on click now but on hover, and it is quite difficult to make it go away. If my mouse is in the incident list browser window, the dropdown is shown. I can only hide it by moving the mouse out of the browser window. (Linux)
> 
> Taking a look. Which browser?

Firefox. It does not happen in Vivaldi.

hmpf:
Ugh. It opens on click the first time in Vivaldi, but afterwards it behaves like firefox.

hmpf (approved):
Works now.

PR #64, opened 2024-08-23
Improve classic details page

hmpf:
This shows everything in the details page that argus classic does, though with no buttons working.

### DIFF
--- a/src/argus_htmx/static/styles.css
+++ b/src/argus_htmx/static/styles.css
@@ -908,6 +908,76 @@ html {
   content: var(--tw-content);
 }
 
+.card {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  border-radius: var(--rounded-box, 1rem);
+}
+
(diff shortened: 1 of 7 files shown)


hmpf:
When submitting any of the buttons I think it would be cleanest if the entire page was reloaded. Because:

* In adding/removing a ticket we change
  1. The badge-type in "Status"
  2. The actual ticket url in the list of incident attributes
  3. The list of events (prepend a change-event)
* In closing/repoening a ticket we change
  1. The badge-type in "Status"
  2. The list of events (prepend a close-event)
  4. The end_time in the list of incident attributes
* In acking an incident we change
  1. The badge-type in "Status"
  2. The list of acks (prepend an ack)

hmpf:
> Some minor polishes like centering and removing some redundant divides might follow in the future, but is approved as is

My first time using tailwind/daisyUI so that is to be expected =)

PR #69, opened 2024-08-26
Select-all incidents feature with update menu placeholder

podliashanyk:


hmpf (approved):
Manually tested. For the impatient: The filterbox now has two tabs, "Filter incidents" and "Update incidents". The latter has placeholder buttons.

PR #83, opened 2024-08-27
Use dialog tag for modals

podliashanyk:
~Depends on #82 (not technically but for being able to test that refactor doesn't break functionality)~


hmpf (approved):
Approved, but move your branch to start off of main, then I'll rebase the details-ack-button branch and the rest on it once it is merged.

hmpf:
No easy way to do that on github it seems.

PR #87, opened 2024-08-28
Show level input with range widget

hmpf:
I'm not happy with the styling. Dunno why it doesn't look as in the daisyUI docs.

### DIFF
--- a/src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
+++ b/src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
@@ -18,7 +18,7 @@
                             {{ field|add_class:"checkbox checkbox-accent border" }}
                         {% elif field.name == "maxlevel" %}
                             <div>
-                                {{ field|add_class:"range range-primary" }}
+                                {{ field|add_class:"range range-primary range-xs" }}
                                 <div class="flex w-full justify-between px-2 text-xs">
                                     {% for tick in "12345" %}
                                     <span>{{ tick }}</span>
--- a/src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
+++ b/src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
@@ -18,13 +18,11 @@
                             {{ field|add_class:"checkbox checkbox-accent border" }}
                         {% elif field.name == "maxlevel" %}
                             <div>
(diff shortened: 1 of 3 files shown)


src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
```diff
@@ -16,6 +16,17 @@
                         <span class="label-text">{{ field.label }}</span>
                         {% if field|field_type == "booleanfield" %}
                             {{ field|add_class:"checkbox checkbox-accent border" }}
+                        {% elif field.name == "maxlevel" %}
```
elfjes, line 19: not sure if this should be the criterium for showing a slider. The cleanest way would be to use a custom widget I think, Then it can be more easily reused for other filter parameters or forms.

but probably after the demo

src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
```diff
@@ -16,6 +16,17 @@
...
+                            <div>
+                                {{ field|add_class:"range" }}
+                                <div class="flex w-full justify-between px-2 text-xs">
```
elfjes, line 22: could benefit from some borders and probably  a `range-primary` class

elfjes (commented):
Looks nice, I like it :). A few comments below though

elfjes:
you say it doesn't look like the docs. This is how it looks for me. looks pretty similar I'd say?:

image

podliashanyk (changes requested):
Tick marks should also be replaced with actual level numbers

src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
hmpf: I mentioned something similar in the newest commit: better to have widget templates (defined in the form widget in the form) than rely too much on widget-tweaks. After the demo :)

hmpf: See #93 for an even better solution.

hmpf:
Tick marks have been replaced with numbers in a hackish, ugly way and I don't think we should do it "properly" until after the demo. See #93.

podliashanyk (approved):
Works for me if we fix #93 after the demo. See a little suggestion below

PR #88, opened 2024-08-28
Make close/reopen buttons work in details page

hmpf:
I'm not sure whether the message should be required or not. It is not required in the data model.

### DIFF
--- a/src/argus_htmx/incidents/forms.py
+++ b/src/argus_htmx/incidents/forms.py
@@ -5,6 +5,6 @@ class AckForm(forms.Form):
     expiration = forms.DateTimeField(required=False)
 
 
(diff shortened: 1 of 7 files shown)


hmpf:
> * I am getting `django.core.exceptions.ValidationError: ['Cannot set a stateless incident as open']` and an error page in the UI when reopening a "stateless" incident

This is what the API does. How should we handle this? Pop up an error message box? Hide the button?

hmpf:
Eh, makes me wish we had tests, this. Works now, and I think tests needs to go on the "when merging with argus proper" todo.

hmpf:
Do you have template debug on in your settings?

PR #91, opened 2024-08-30
Make theme selectable via dropdown in user menu

### LINKED ISSUE #75: Prettify theme-page
- [ ] Mark the theme names as clickable
- [ ] Add some air

Maybe replace with a dropdown?

podliashanyk:
Closes #75 

Do we need to keep the old theme page?


hmpf (commented):
Will test manually

hmpf:
We do not need the old theme-page.

PR #92, opened 2024-09-02
Use minified HTMX

### LINKED ISSUE #70: Store HTMx files locally, do not load from CDN
.. we need to be able to demo the site even if the network is flaky...

- [x] Store HTMx-files locally
- [x] Document what to do to upgrade the HTMx-files to a new version

podliashanyk:
Closes #70


src/argus_htmx/templates/htmx_base.html
```diff
@@ -6,6 +6,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% block title %}<title>Argus Server: {{ page_title }}</title>{% endblock %}
 
+    <script src="{% static "htmx.min.js" %}"></script>
+    <!-- HTMX Config -->
+    <meta name="htmx-config" content='{"useTemplateFragments":"true"}'>
```
elfjes, line 11: Do we need to use this? It doesn't seem like we use `<template>` tags anywhere

elfjes (commented):
nice, but with a comment

docs/development/dependencies.rst
```diff
@@ -0,0 +1,24 @@
...
+
+1. **Backup Current File**:
+   Ensure you have a backup of the current `htmx-*.min.js` file located in the `static` directory.
+
+2. **Download the Latest Version**:
+   Obtain the latest version of minified HTMX from the `UNPKG`_, for example:
+   `https://unpkg.com/htmx.org@2.0.2/dist/htmx.min.js`
+
+3. **Replace Old File**:
+   Replace the old HTMX file in your project with the new file from the latest version. This involves updating the `htmx-*.min.js` file in the `static` directory. Make sure to replace the `*` in the file name with the new version number.
```
elfjes, line 14: Nitpick: this could make a great Make command

src/argus_htmx/templates/htmx_base.html
```diff
@@ -5,7 +5,10 @@
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     {% block title %}<title>Argus Server: {{ page_title }}</title>{% endblock %}
-
+    <!-- HTMX -->
+    <script src="{% static "htmx-2.0.2.min.js" %}"></script>
```
elfjes, line 9: You know what would be cool, a template filter that checks the current static directory for htmx files and picks the latest. something with `pathlib.Path.glob` together with  a `version_info` namedtuple or so. slap a `@functools.cache` on it for performance

elfjes (commented):
Some nice-to-haves:

docs/development/dependencies.rst
hmpf: The version-number wrangling would make it quite clunky.

src/argus_htmx/templates/htmx_base.html
hmpf: Thanks to having to take the type of storage in use into account (filenames can be rewritten) this is a big enough change that it deserves its on issue.

docs/development/dependencies.rst
hmpf: We could have a management command that does this, which could be run from tox/make.

hmpf:
Frankly there's a bit of clunk involved with storing the version in the filename. We ought to store the version *somewhere* though. Hm.

src/argus_htmx/templates/htmx_base.html
hmpf: It's copied from the original though so nothing changes. I'm curious myself, what is this for?

src/argus_htmx/templates/htmx_base.html
elfjes: We don't neccesarily need to keep old versions I think, but it makes the aforementioned Make command easier if the template has some htmx version discovery. The make command can then delete old versions, add the new one, and Django automatically picks up on it

PR #96, opened 2024-09-02
Support extra widget through context processor

elfjes:
Supports adding a widget to the incident listing through settting the `incidents_extra_widget` context variable to a template path that will then be included next to the menu bar. The widget will take up 1/3 of the width.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -250,3 +250,5 @@ injects a ``incidents_extra_widget`` variable that points to a html template::
         return {
             "incidents_extra_widget": "path/to/_extra_widget.html",
         }
+
+*note* don't forget to include the context processor in your settings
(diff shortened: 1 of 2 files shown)


hmpf (changes requested):
Works.

Please include an example template (which has lorem or a link to rick roll or something) to exemplify styling. Either in the docs or as a file together with the other partial templates, for instance `_incident_extra_widget_example.html`

elfjes:
> Works.
> 
> Please include an example template (which has lorem or a link to rick roll or something) to exemplify styling. Either in the docs or as a file together with the other partial templates, for instance `_incident_extra_widget_example.html`

Done. I have added it in the docs, so as not to create any dangling files

hmpf (approved):

PR #98, opened 2024-09-03
Give timestamp a minimum width in incident list

hmpf:


elfjes (approved):
Given that this is the exact same implementation as https://github.com/GEANT/geant-argus/pull/30 I can't not approve.

But maybe there's some room for discussion. Ideally you'd like to set the class on the `<th>` element directly. that way you can also set relative width for columns. I can think of two implementation to achieve that:

1. let `IncidentTableColumn.header_template` include the `<th>` tag
2. Introduce a `IncidentTableColumn.extra_header_classes` variable that adds class to the `<th>`

I think (1) may be the simplest

hmpf:
Not before the demo. This is a deeper issue probably, how much should be styled in python and how much with daisy/tailwind directly.

PR #106, opened 2024-09-03
Improve "how to run argus_htmx" in the readme

hmpf:


johannaengland (commented):
For now I can only approve the part using the `Django-style configuration`. That worked great, the only thing that I had to search for, what I would like mentioned in the docs is that one can log in at `accounts/login`, because I tried to go via the Django admin and it would not let me

hmpf:
You should be forced to login if you try to visit `/incidents/` actually. Could you log out and try that?

johannaengland:
> You should be forced to login if you try to visit `/incidents/` actually. Could you log out and try that?

I will get forwarded to `http://localhost:8000/login/?next=/incidents/` and get a `Page not found` error

hmpf:
hah, found it.. new commit on its way.

johannaengland:
Can confirm that this fixes my problem

PR #107, opened 2024-09-03
Add HTMX-aware messages-middleware

hmpf:
Copy (django.contrib) messages list to response header so HTMx can easily access them.


lunkwill42 (approved):
I just see a lot more legacy code here, but that seems to be all the rage in argus-htmx ;-)

PR #112, opened 2024-09-04
Make a column for tickets

### LINKED ISSUE #114: Make a column for tickets
Closed by #112

hmpf:
Closes #114

image


johannaengland (changes requested):
Without any extra customization the contrast is really bad: 
image

And if I click on the link having entered the ticket url "example.com" previously it wants to send me to "http://localhost:8000/incidents/example.com" instead

hmpf:
If you do not see the icon as teal in dark mode have you tried reloading the page or rebuilding the styles?

(By: stand in `src/argus_htmx/tailwindtheme`, run `tailwindcss -c tailwind.config.js -i ./styles.css --output ../static/styles.css`, reload page.)

hmpf:
Also, absolute urls should be stored with http or https so this is a problem with validating form input and irrelevant to this PR. I note that the classic frontend do complain if http/https is missing. The admin does not, and the new frontend does not. I'll make issues.

PR #120, opened 2024-09-05
Hide source type in source column in incident list

### LINKED ISSUE #58: Fix the source column
In the "source" column, do one of:

1. only show source, not source + (source type)
2. use an icon for source type

1 is a one-liner.

johannaengland:
Closes #58.

image


elfjes (commented):
Fine by me, we're not using this column anyway :)

hmpf (approved):
Possibly the easiest thing ever?

PR #122, opened 2024-09-05
Style status and ack columns

### LINKED ISSUE #99: Use icons for open/closed, acked/unacked (or show nothing for unacked) in their respective columns, like it is in classic

hmpf:
Closes #99 

image


hmpf:
This will not be used for the demo, just kept in the ready-made columns as an example.

PR #129, opened 2024-09-06
Show "Ticket" instead of "Acknowledged" in status block of details page

### LINKED ISSUE #128: Status part of details page says "Acknowledged" for incident that has a ticket
Instead of stating that an incident has a ticket it instead just states that this incident is acknowledged, even for incidents that don't have any related acknowledgements. 

image

johannaengland:
Closes #128. 

Open for discussion about which of the commits to merge - in the first two commits I made it look like the original Argus-frontend and in the third I added that clicking the link will open it in another tab.


hmpf:
I think opening in new tab for external pages is always correct.

PR #136, opened 2024-09-09
Use ruff to format/lint the codebase

johannaengland:
It is not possible yet to use the ruff formatter in MegaLinter, therefore I have set up to do the formatting and linting using ruff directly.

This also removes use of flake8 in pre-commit. 

When this PR is approved I will squash all the fixing of the linting rules and the formatting to one commit and then add this commit to `.git-blame-ignore-revs`.

Also then I can change the workflow to only format/lint the changed files.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,2 @@
-# Format codebase using ruff @johannaengland 09/09/2024
-07fcc0fa89e719a7cd00d36a3f12e3b552ac6ef2
+# Format all python with ruff
+47f713ddce20e3c751e3cbd05a8671dda8f9a054
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Format codebase using ruff @johannaengland 09/09/2024
+07fcc0fa89e719a7cd00d36a3f12e3b552ac6ef2
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,21 @@
+name: Ruff
+on:
+  push:
+    branches: main
+  pull_request:
+
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Lint
+        uses: chartboost/ruff-action@v1
+        with:
+          args: "check"
+          # changed-files: 'true'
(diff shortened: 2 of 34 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Format codebase using ruff @johannaengland 09/09/2024
+07fcc0fa89e719a7cd00d36a3f12e3b552ac6ef2
```
hmpf, line 2: Code is added to this repo by squashing so this will not be correct.

hmpf (changes requested):
This needs splitting into at least two PR's, isort adds much noise. I'll spread this into several direct pushes and then this PR can be rebased to add the github actions.

Btw, have you remembered to set up pre-commit? It's already configured. There are newlines missing from several files.

elfjes (approved):
I like it. the ignore-rev can be fixed by
1) squash all non-formatting commits
2) format everything in 1 commit
3) add that commit to `.git-blame-ignore-revs`
4) force push and FF-merge to main

elfjes:
Also: next step would be linting/formatting of templates (djlint?) and perhaps js/css?

johannaengland:
> Also: next step would be linting/formatting of templates (djlint?) and perhaps js/css?

#137

.git-blame-ignore-revs
johannaengland: Yes, I am aware, I was not planning on merging it like this, I just added this commit to not forget about it later

hmpf:
I have pushed the ruff-ed code to main with the correct hash in .git-blame-ignore-revs.

I have not removed the flake8 rule in pre-commit, and I am not linting in pre-commit right now.

I would like to see one PR that sets up github actions for this. Let's wait with isort, it has a *very' low priority and I Have Opinions™, I want to play with isort config options and how they interact with ruff before I let it loose on anything.

hmpf:
We should have a changelog on this somewhere as it is very relevant for developers. I have found two handy flags when using "git blame":

```
[blame]
        markIgnoredLines = true
        markUnblamableLines = true
```

Try it and see with one of the files that has been ruffed.

I like what they do, they should be mentioned in the changelog but not in notes.

johannaengland:
This has been mostly replaced by #176 and #175. Other PRs to bit by bit introduce more linting rules might follow soon.

PR #137, opened 2024-09-12
Add linting of CSS and HTML files using MegaLinter

### LINKED ISSUE #40: Linting/Code formatting
I noticed some style inconsistencies in the code (eg inconsistent single/double quotes in `incident.views.incident_add_ack` and was wondering if we should set up some code formatting (ie. black)? Always good to start using a formatter as early as possible so that fewer code needs to be touched once set up.

Also it might be good to not only autoformat python code but also the templates? I like djlint for that.

johannaengland:
Closes #40.

djLint will be used to format and lint HTML files and stylelint to lint CSS files. I have not added formatting of CSS files since that is not supported in MegaLinter as far as I found out and otherwise I am too unsure what to use to format CSS files. 

MegaLinter is currently failing because linting some HTML files produces errors. I would like some help to fix these errors or to tell me that these rules can be ignored.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Reformat html files using djLint @johannaengland 12/09/2024
+c53dafbd8632f101d960a67d950245e4bb42f21e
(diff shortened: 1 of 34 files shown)


johannaengland:
I just saw that only one CSS file has been linted - `src/argus_htmx/tailwindtheme/styles.css`, because `src/argus_htmx/static/styles.css` is in `.gitignore`. Which CSS files should be actually linted?

johannaengland:
We have decided to not format CSS files and use djLint by itself to lint/format HTML files

Issue #59, opened 2024-08-22
Move everything in ./src/argus_htmx/tailwindtheme to ./tailwind


hmpf:
We only need this stuff when developing, not in production, so move it out of `src/`.

Remember to update the README!

PR #138, opened 2024-09-17
Add path for finding content under current dir

stveit:
Previously required you to stand a step deeper than `src/argus_htmx` in order for tailwind to find html content (the path did a `../templates`, so you had to go a step back then into templates)

This PR find all templates under the current dir, so should work for repo root and anywhere above templates


hmpf:
The technically correct solution is probably #59 (and adding running the command to both tox and make..)

hmpf (approved):

PR #145, opened 2024-09-26
Selectable incident list page size

### LINKED ISSUE #76: It should be possible to select how many incidents to show per page

elfjes:
Closes #76

Add a select element for choosing the page size (max number of incident rows shown in the table). The default page size is 10 and selectable page sizes are [10, 20, 50, 100]. It is possible to override these numbers by setting the ``ARGUS_INCIDENTS_DEFAULT_PAGE_SIZE`` and ``ARGUS_INCIDENTS_PAGE_SIZES`` setting
respectively.

some noteworthy things
* hitting enter on a filterbox fields would trigger a page reload (something djang-forms maybe?). it was necessary to intercept the enter event in order to include the page_size form value, otherwise it would only send in the filter box params
* I wrapped the whole refresh-info stats element in a form, because I think we also want to make the auto-refresh frequency selectable

This PR is still a wip because I'm not quite happy with the current styling of the refresh-info, since the page-size selector increased the height. If any of you has a good styling idea, please go ahead :)

### DIFF
--- a/src/argus_htmx/incidents/views.py
+++ b/src/argus_htmx/incidents/views.py
@@ -218,7 +218,7 @@ def incident_list(request: HtmxHttpRequest) -> HttpResponse:
         "base": base_template,
         "page": page,
         "last_refreshed": last_refreshed,
-        "update_interval": 3000,
+        "update_interval": 30,
         "page_size": page_size,
         "all_page_sizes": ALLOWED_PAGE_SIZES
     }
--- a/src/argus_htmx/incidents/views.py
+++ b/src/argus_htmx/incidents/views.py
@@ -4,6 +4,7 @@ import logging
 from typing import Optional
 from datetime import datetime
 
+from django.conf import settings
 from django.contrib.auth.models import Group
 from django.contrib.auth import get_user_model
 from django.contrib import messages
@@ -26,7 +27,8 @@ from .forms import AckForm, DescriptionOptionalForm, EditTicketUrlForm, AddTicke
 
 User = get_user_model()
 LOG = logging.getLogger(__name__)
-
+DEFAULT_PAGE_SIZE = getattr(settings, "ARGUS_INCIDENTS_DEFAULT_PAGE_SIZE", 10)
+ALLOWED_PAGE_SIZES = getattr(settings, "ARGUS_INCIDENTS_PAGE_SIZES", [10, 20, 50, 100])
 
 def prefetch_incident_daughters():
     return (
@@ -171,6 +173,14 @@ def incident_detail_edit_ticket(request, pk: int):
     return redirect("htmx:incident-detail", pk=pk)
 
 
+def _get_page_size(params):
+    try:
+        if (page_size:=int(params.pop('page_size', DEFAULT_PAGE_SIZE))) in ALLOWED_PAGE_SIZES:
+            return page_size
+    except ValueError:
+        pass
+    return DEFAULT_PAGE_SIZE
+
 @require_GET
 def incident_list(request: HtmxHttpRequest) -> HttpResponse:
     columns = get_incident_table_columns()
@@ -188,8 +198,8 @@ def incident_list(request: HtmxHttpRequest) -> HttpResponse:
 
     # Standard Django pagination
     page_num = params.pop("page", "1")
-    PAGE_SIZE = 10
-    paginator = Paginator(object_list=qs, per_page=PAGE_SIZE)
+    page_size = _get_page_size(params)
+    paginator = Paginator(object_list=qs, per_page=page_size)
     page = paginator.get_page(page_num)
 
     # The htmx magic - use a different, minimal base template for htmx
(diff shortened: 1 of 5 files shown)


src/argus_htmx/incidents/views.py
```diff
@@ -208,8 +218,9 @@ def incident_list(request: HtmxHttpRequest) -> HttpResponse:
         "base": base_template,
         "page": page,
         "last_refreshed": last_refreshed,
-        "update_interval": 30,
-        "per_page": PAGE_SIZE,
+        "update_interval": 3000,
```
hmpf, line 221: Why do you change `update_interval`? (Should probably also be a setting, though.)

src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
```diff
@@ -1,13 +1,13 @@
...
-    >
+<form id="incident-filter-box"
+      hx-get="{% url 'htmx:incident-list' %}"
+      hx-include="#table-refresh-info, #incident-filter-box"
+      hx-trigger="keydown[keyCode==13], change delay:100ms"
+      hx-target="#table"
+      hx-swap="outerHTML"
+      hx-push-url="true"
+      onkeydown="if (event.keyCode === 13) event.preventDefault();">
+    <fieldset>
```
hmpf, line 10: You change to post on enter, not after delay? What's the `onkeydown` for?

src/argus_htmx/templates/htmx/incidents/_incidents_refresh_info.html
```diff
@@ -1,22 +1,30 @@
...
+            <dd class="stat-value text-2xl text-base"> {{ filtered_count }}</dd>
+        </div>
+        <div class="stat py-2">
+            <dt class="stat-title text-neutral-content/80">Per page</dt>
+            <dd class="stat-value text-2xl text-primary">
+                <select class="select select-xs" name="page_size">
+                    {% for ps in all_page_sizes %}
+                        <option value="{{ ps }}" {% if ps == page_size %}selected{% endif %}>{{ ps }}</option>
+                    {% endfor %}
+                </select>
```
hmpf, line 18: This could be encapsulated in a django form.

(Really need to move up to Django 5.1 for more form goodness first though.)

hmpf (commented):
See questions.

I'll test this manually...

src/argus_htmx/incidents/views.py
elfjes: haha oops, that's a leftover from my development flow. I often change it to a very high value so that I can properly inspect the DOM without interruptions (one reason why I'd like the update interval configurable)

src/argus_htmx/templates/htmx/incidents/_incident_filterbox.html
elfjes: So it used to be only `change delay` for htmx. But hitting enter while one of the form fields were focused also caused a get request to be fired. This was not an ajax request, and not triggered by htmx. I'm not quite sure what the origin of the request is. However, that request would only add the filterbox form fields to the query parameters, but we also need to include the current page size. So I had to prevent the keydown enter event, and add it as and hx-trigger. hx-trigger now has both triggers

If you can figure out a better solution, please feel free :D

src/argus_htmx/templates/htmx/incidents/_incidents_refresh_info.html
elfjes: Sure, but I find it sometimes hard to customize django forms. And since it was one select, it seemed ok to not do it as a form.

but feel free to change :)

src/argus_htmx/incidents/views.py
elfjes: anyway I've reverted it

src/argus_htmx/templates/htmx/incidents/_incidents_refresh_info.html
hmpf: Had it been before the nordunet conference I would have said "after-demo". No rush :)

hmpf:
Needs fixed: The page size is not conserved when changing pages.

hmpf:
The styling needs some love, too much padding around the value:

image

hmpf:
(After a round of "make tailwind" (that fix was sooo worth it)):

image

Better but could still need some love.

elfjes:
> (After a round of "make tailwind" (that fix was sooo worth it)):
> 
> image
> 
> Better but could still need some love.

Yup, which was the reason for my last comment in the OP

hmpf:
> > image
> 
> /../ I'm not sure how to improve it

Have the same font-size for all the numbers and center all the numbers vertically I guess, not that I know the daisy/tailwind-magic for that.

hmpf:
I'll approve it when the page-switching bug is fixed, the styling polish can live on in a new polish issue.

elfjes:
Ok, styling fixed: 
image

elfjes:
> Needs fixed: The page size is not conserved when changing pages.

fixed

hmpf (approved):
Tested and approved.

PR #146, opened 2024-09-27
Show different "no hits" text for empty db

### LINKED ISSUE #105: List empty due to no incidents matching the filter should have different text than empty due to empty database
When the list is empty due to no incidents matching the filter, the text that replaces the table should say just that: "No incidents match the current set filter" or something like that. Different from when we have zero incidents in the database.

hmpf:
Closes #105


elfjes (approved):
not a 100% sure about the wording `No incidents on record` but I don't have  a better suggestion 😅

hmpf:
It'll only be shown on first install before glue services have been set up, or if something dramatic has happened to the contents of the database, so if we wanted I guess we could even hide an easter egg there =)

PR #159, opened 2024-10-03
Discoverability of tailwind css snippets

### LINKED ISSUE #157: Add discoverability of custom css files/snippets for tailwind integration
#156 adds additional css. We (geant) need to be able to pick up on this and combine it with our own custom css. I see two options:

1) Argus htmx provides discoverability:
    * extend the `HtmxFrontendConfig` to add a variable that points to one or more css files (we need to be able to import something that has a `__file__` attribute so that we know the (relative) path where to look for the css files)
    * we templatize our `styles.css`
    * in our `tailwind_config` command, we read this variable and include the files in our templated `styles.css`
 2) Invert the dependency
    * extend argus_htmx `tailwind_config` to support reading a list of additional css files from a setting (and additional features we need)
    * we supply the additional settings (which includes our css customizations) and subsequently use argus_htmx `taiwind_config` instead of our own

I'm leaning mostly towards 2, but I'm not fully sure yet how discoverability is going to work out

elfjes:
fixes #157 

TODO: 
- [x] formatting of `{{ daisyuithemes }}`
- [x] some cleanup on all the iterators and generators


src/argus_htmx/apps.py
```diff
@@ -1,7 +1,11 @@
...
 from django.apps import AppConfig
 
 
 class HtmxFrontendConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     label = "argus_htmx"
     name = "argus_htmx"
+
+    def tailwind_css_files(self):
+        yield from pathlib.Path(__file__).parent.glob("tailwindtheme/snippets/*.css")
```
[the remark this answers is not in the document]
elfjes: It could be, technically, but since it's the reference/example implementation, keeping it as an instance method conveys to third party developers that it can be either an instance method, classmethod or staticmethod. Making it a staticmethod loses that. For me it was non-obvious whether django instantiates the appconfig (and thus we can use normal methods) or uses the class directly, until I tried it out.

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -1,8 +1,13 @@
+import inspect
```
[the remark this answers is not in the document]
elfjes: repo could benefit from more linting :D

src/argus_htmx/tailwindtheme/snippets/20-sample.css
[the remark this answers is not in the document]
elfjes: sure :). we'll have a real file in there soon enough

src/argus_htmx/tailwindtheme/tailwind.config.js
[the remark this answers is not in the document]
elfjes: So, my opinion is that if you want to commit this file, it should be comitted as it is generated from the latest HEAD. (similar to static/styles.css) so that there can be no ambiguity as to how the file was created. It's also why I spent some effort to make it produce formatted code, so that you don't need any (manual) post-processing to commit a nicely looking file

src/argus_htmx/settings.py
```diff
@@ -16,8 +16,10 @@
 ]
 
 # Tailwind config template relative to the repository root directory
-TAILWIND_CONFIG_TEMPLATE = "src/argus_htmx/tailwindtheme/tailwind.config.template.js"
-TAILWIND_CONFIG_TARGET = "src/argus_htmx/tailwindtheme/tailwind.config.js"
+TAILWIND_CONFIG_TEMPLATE = "tailwind/tailwind.config.js"
+TAILWIND_CSS_TEMPLATE = "tailwind/styles.css"
+TAILWIND_CONFIG_TARGET = 'src/argus_htmx/tailwindtheme/tailwind.config.js'
+TAILWIND_CSS_TARGET = 'src/argus_htmx/tailwindtheme/styles.css'
```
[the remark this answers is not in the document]
elfjes: Yeah that makes sense

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -12,38 +17,111 @@
...
+            "themeoverride": getattr(
+                settings,
+                "TAILWIND_THEME_OVERRIDE",
+                argus_htmx_settings.TAILWIND_THEME_OVERRIDE,
+            ),
+            "daisyuithemes": textwrap.indent(
+                json.dumps(get_themes(), indent=2),
+                prefix=10 * " ",
+                predicate=lambda line: line != "[\n", # this is kinda hacky, but eh
+            ),
```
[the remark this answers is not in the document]
elfjes: Yes, outputting get_themes() directly produces a single line of json in the tailwind config. This doesn't look nice for a file that is committed (see earlier comment). With this it looks decent. Perhaps if we have some more linting steps, we could reformat the file after creation and it would not be necessary,

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -12,38 +15,107 @@
 # Copied from https://github.com/GEANT/geant-argus/pull/15 with minor modifications
 class Command(BaseCommand):
     help = """\
-    
-Uses the 
-template specified in the TAILWIND_CONFIG_TEMPLATE setting (default: tailwind.config.js)
-to dynamically build a tailwind.config.js. The template should contain:
- - a'{{ daisyuithemes }}' section without square brackets that will be popuplated by the 
+Uses the template specified in the TAILWIND_CONFIG_TEMPLATE setting (default: tailwind/tailwind.config.js)
```
hmpf, line 18: Where's the root? Are we at repo root or somewhere in the code? What are we relative *to*?

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -12,38 +15,107 @@
...
         pathlib.Path(target_path).write_text(
-            self.render_config(template_path=template_path, context=context)
+            self.render_file(template_name=template_name, context=context)
         )
 
-        self.stdout.write(f"Wrote tailwind config to '{target_path}'")
+        self.stdout.write(f"Wrote {name} to '{target_path}'")
+
+    @staticmethod
+    def render_file(template_name: str, context):
```
hmpf, line 100: render_*file* might not be the best name. You're not writing the actual file, you're rendering a *string* that will be used by something else. Django would just call this `render`, other would probably call it `dumps`, for dump string.

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -12,38 +15,107 @@
...
+                self.make_relative(p, target_dir)
+                for p in sorted(self.get_css_files(), key=lambda p: p.stem)
+            ),
         }
-        template_path = pathlib.Path(template_loc)
-        if not template_path.is_file():
-            self.stdout.write(f"{template_loc} is not a file")
-            return
 
+    def write_template(self, template_name, target_path, context, name):
```
hmpf, line 92: You're not writing a template but the result of a rendered template. `write_file` is better.

src/argus_htmx/management/commands/tailwind_config.py
```diff
@@ -12,38 +15,107 @@
...
+
+    @staticmethod
+    def render_file(template_name: str, context):
+        template = get_template(template_name)
+        return template.template.render(make_context(context, autoescape=False))
+
+    @staticmethod
+    def get_template_dirs():
+        for engine in engines.all():
+            yield from getattr(engine, "template_dirs", [])
```
hmpf, line 107: I need to check Django source code for this.

src/argus_htmx/tailwindtheme/README.md
```diff
@@ -0,0 +1,6 @@
+# Files for tailwindcss
+
+`styles.css` and `tailwind.config.js` are automatically generated by the `tailwind_config` django
+managemet command. The source for these files are in `../templates/tailwind/`
```
hmpf, line 4: s/managemet/management/

The .. is relative to the README? More specificity would be nice.

src/argus_htmx/tailwindtheme/styles.css
```diff
@@ -1,3 +1,2 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import 'snippets/10-tailwind.css';
+@import 'snippets/20-sample.css';
```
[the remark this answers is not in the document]
hmpf: It's a pretty standard trick in unix.

src/argus_htmx/management/commands/tailwind_config.py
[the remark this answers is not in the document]
hmpf: I agree that we indent everything, including the result of `get_themes()`

hmpf (changes requested):
Please improve some names, also found a typo.

I will need to have a look at the engines-magic.

src/argus_htmx/management/commands/tailwind_config.py
elfjes: It's a template path, so it's whatever template is found by the template engine with that path

src/argus_htmx/management/commands/tailwind_config.py
elfjes: that would be in `django.template.backends.base.BaseEngine.template_dirs`

src/argus_htmx/management/commands/tailwind_config.py
hmpf: Could you add that info?

elfjes: done

hmpf (approved):

src/argus_htmx/management/commands/tailwind_config.py
hmpf: It depends on the TEMPLATES-setting so all good.

hmpf (approved):

Issue #154, opened 2024-09-27
clicking while the modal is open seems to flash something. clicking and holding make this more apparent, I don't think it should do that:


hmpf:
clicking while the modal is open seems to flash something. clicking and holding make this more apparent, I don't think it should do that:

![image](https://github.com/user-attachments/assets/185f1740-3e23-4fbd-9d60-87e0a9f664b4)

_Originally posted by @elfjes in https://github.com/Uninett/argus-htmx-frontend/pull/74#pullrequestreview-2315015785_

PR #162, opened 2024-10-08
Fix broken layout of bulk modals

### LINKED ISSUE #152: Fix the positioning of the bulk modals
> RFC:
> 
> - Not a fan of the placement of bulk modals in top left corner. For the reference the modals in incident deyails view are nice and centered, while all bulk modals show up like this to me:
> Screenshot 2024-09-19 at 11 06 43

_Originally posted by @podliashanyk in https://github.com/Uninett/argus-htmx-frontend/pull/74#pullrequestreview-2314933891_

podliashanyk:
Closes #152, #154

The problem was due to conflicting grid layout styles between `<dialog>` (the modal box) and `.menu li` (listitem tags that modals were wrapped with).
There are alternative solutions to the one in this PR:

- break the overlap between the conflicting styles by adding another tag (layer) between listitem and modal. Achievable by adding a wrap `<div class="contents">...</div>` around modal (either directly in base template `_base_form_modal.html`, or around each `include` in `_incidents_update_menu.html`). Docs about display option contents: https://tailwindcss.com/docs/display#contents. Pros: Very strong alternative, plus it could somewhat be a failsafe against future broken placements of modals. Cons: Adds "empty" layer of nesting, aka adding structural tag purely for styling purposes
- add targeted CSS overrides and test that they work properly with existing styles

### DIFF
--- a/src/argus_htmx/templates/htmx/incidents/_incidents_update_menu.html
+++ b/src/argus_htmx/templates/htmx/incidents/_incidents_update_menu.html
@@ -1,8 +1,8 @@
-<ul class="menu menu-horizontal gap-2">
+<div class="menu menu-horizontal gap-2">
     {% block update_menu_content %}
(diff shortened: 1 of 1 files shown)


hmpf (approved):
Let's get this out of the way.

Issue #127, opened 2024-09-06
We need to organize the templates better


hmpf:
We should use namespacing so that template fragments for modals are in one directory, templates for incident columns in another etc.

hmpf:
I was thinking:

```
incidents/
|- list/
|  |- modals/..
|  |- columns/..
|  \ - incident_list.html
\ detail/
   |- modals/..
    \ - incident_detail.html
```

.. considering that bulk modal and single incident modals need different texts thanks to English plural vs. singular nouns in help texts and titles etc.

elfjes:
That sounds good to me. I don't have any strong feelings towards how to organize, but some improved organization sounds good. We can match whatever organization will be settled on :)

PR #164, opened 2024-10-09
Add endpoints for updating incidents in bulk

podliashanyk:
Closes Uninett/argus-htmx-frontend#153 

Waiting for Uninett/argus-htmx-frontend#80 for more effective implementation of bulk updates.

Using `hx-include` [(docs)](https://htmx.org/attributes/hx-include/) for including info about what incidents are selected when posting bulk update.

### Will be fixed in separate PRs:
- Uninett/argus-htmx-frontend#178 
- Uninett/Argus#1058

### DIFF
--- a/src/argus_htmx/incidents/views.py
+++ b/src/argus_htmx/incidents/views.py
@@ -144,7 +144,7 @@ def incidents_update(request: HtmxHttpRequest):
     try:
         formclass, callback_func = INCIDENT_UPDATE_ACTIONS[form_name]
     except KeyError:
-        LOG.error(f"Unrecognized form name {form_name} when updating incidents.")
+        LOG.error("Unrecognized form name %s when updating incidents.", form_name)
         return HttpResponseBadRequest("Invalid form name")
     formdata, incident_ids = get_form_data(request, formclass)
     if formdata:


src/argus_htmx/incidents/urls.py
```diff
@@ -13,4 +13,5 @@
     path("<int:pk>/reopen-detail/", views.incident_detail_reopen, name="incident-detail-reopen"),
     path("<int:pk>/ticket-edit-detail/", views.incident_detail_edit_ticket, name="incident-detail-edit-ticket"),
     path("<int:pk>/ticket-add-detail/", views.incident_detail_add_ticket, name="incident-detail-add-ticket"),
+    path("bulk-ack/", views.incidents_bulk_ack, name="incidens-bulk-ack"),
```
elfjes, line 16: typo
```suggestion
    path("bulk-ack/", views.incidents_bulk_ack, name="incidents-bulk-ack"),
```

src/argus_htmx/templates/htmx/incidents/_incidents_update_menu.html
```diff
@@ -1,7 +1,7 @@
 <div class="menu menu-horizontal gap-2">
     {% block update_menu_content %}
         {% include 'htmx/incidents/_incident_ticket_edit_modal.html' with dialog_id="add-ticket-dialog" button_class="btn-accent" button_title="Change tickets" header="Change tickets" explanation="Write an URL of an existing ticket, or nothing to remove existing ticket urls" cancel_text="Cancel" submit_text="Submit" %}
-        {% include 'htmx/incidents/_incident_acknowledge_modal.html' with dialog_id="create-acknowledgment-dialog" button_class="btn-accent" button_title="Acknowledge" header="Submit acknowledgment" explanation="Write a message describing why these incidents were acknowledged" cancel_text="Cancel" submit_text="Submit" %}
+        {% include 'htmx/incidents/_incident_acknowledge_modal.html' with dialog_id="create-acknowledgment-dialog" endpoint="htmx:incidens-bulk-ack" button_class="btn-accent" button_title="Acknowledge" header="Submit acknowledgment" explanation="Write a message describing why these incidents were acknowledged" cancel_text="Cancel" submit_text="Submit" %}
```
elfjes: typo
```suggestion
        {% include 'htmx/incidents/_incident_acknowledge_modal.html' with dialog_id="create-acknowledgment-dialog" endpoint="htmx:incidents-bulk-ack" button_class="btn-accent" button_title="Acknowledge" header="Submit acknowledgment" explanation="Write a message describing why these incidents were acknowledged" cancel_text="Cancel" submit_text="Submit" %}
```

src/argus_htmx/templates/htmx/_base_form_modal.html
```diff
@@ -6,7 +6,10 @@
 <dialog id="{{ dialog_id }}" class="modal">
     <div class="modal-box card card-compact shadow-xl">
         <div class="divider divider-start"><h3 class="card-title">{{ header }}</h3></div>
-        <form id="{{ dialog_id }}-form" class="card-body" method="post" action="{{ endpoint }}">
+        <form id="{{ dialog_id }}-form" class="card-body"
+              hx-post="{% url endpoint|default:'htmx:incident-list' %}"
+              hx-include="[name='selected_incidents']"
+        >
```
elfjes: This file is called base_form_modal. Does it make sense to add the hx-include in here? Perhaps rename to base_incident_update_modal? Or, if this modal is used elsewhere for not updating incidents, make it something else?

elfjes (changes requested):
Perhaps it's better to create an `incidents/update` endpoint that takes in an action parameter so that you don't need additional endpoints for different actions and based on the length of `selected-incidents` can work for one or more selected incidents

src/argus_htmx/templates/htmx/_base_form_modal.html
hmpf: There's also #127

hmpf (commented):
"/bulk/ack/"  would look prettier. I'll look over #80 again.

src/argus_htmx/templates/htmx/_base_form_modal.html
elfjes: We have some other modals that we use to display some data:
https://github.com/GEANT/geant-argus/blob/develop/src/geant_argus/geant_argus/templates/htmx/incidents/_incident_quick_glance_modal.html
https://github.com/GEANT/geant-argus/blob/develop/src/geant_argus/geant_argus/templates/htmx/incident_details/event_properties.html
But no need to make a generic component out of it

We may want to look at extending base_form_modal for our own custom incident bulk edit modals. 

But the main point I wanted to make here is that base_form_modal.html is used in other places in argus_htmx and you don't necessarily want to hx-include selected_incidents there

hmpf:
> Perhaps it's better to create an `incidents/update` endpoint that takes in an action parameter so that you don't need additional endpoints for different actions and based on the length of `selected-incidents` can work for one or more selected incidents

When #80 is in there's just a matter of selecting the correct bulk update function to use. Whether it is easier to map one endpoint to one bulk update function or have a hidden bulk-type-field in the form is eminently bikesheddable though security-wise I would prefer avoiding the hidden field.

Also, we should have an actual form to validate each modal. Later on I'd prefer if we rendered that form too, then we might not need one template per modal since we'd pass in the form as context.

elfjes:
I don't see why `bulk/ack` would be less a security issue than `bulk/?action=ack`?

One argument for having an action parameter, is that, with a bit of addional work, we can easily register our own actions with a callback, and we won't need to add much overhead to support our custom actions

src/argus_htmx/templates/htmx/_base_form_modal.html
elfjes: sure, but I think it should read
```
            {% else %}
          method="post"
          action="{% url endpoint|default:'htmx:incident-list' %}"
            {% endif %}
```

hmpf:
We should use forms instead of having one separate template per type, yes. Polish. We could make our own "generic view" for these, a function or class that takes a form and the context that is now explicit in the template via the include. Polish. (Class would probably be prettier. Subclass it, set the form and helptext and bulk function (and with Django 5.1: form-template, and register the url to be anything you want. Done.)

I would very much like to have us running on Django 5.1 first (and having abandoned 4.2).

I would also like an example of bulk actions you need in case they need extra parameters. To make something generic we need examples.

@podliashanyk finish this up and make a new issue for the very-much-like-to-haves.

src/argus_htmx/incidents/views.py
```diff
@@ -110,6 +121,43 @@ def incident_detail_add_ack(request, pk: int, group: Optional[str] = None):
     return redirect("htmx:incident-detail", pk=pk)
 
 
+def user_is_group_member(user: User, group_name: str):
+    group = get_object_or_404(Group, name=group_name)
+    is_group_member = user.groups.filter(pk=group.pk).exists()
+    if group and not is_group_member:
+        raise PermissionDenied("User {request.user} is not a member of the correct group")
+    return is_group_member
```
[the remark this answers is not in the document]
hmpf: Also, have you tried `group in user.groups.all()`?

Hm, I think I'll modify the User-class myself.

src/argus_htmx/templates/htmx/incidents/_base_incident_update_modal.html
```diff
@@ -0,0 +1,11 @@
+{% extends 'htmx/_base_form_modal.html' %}
+
+{% block form_control %}
+  {% if action_type == "bulk-update" %}
+    hx-post="{% url endpoint|default:'htmx:incident-list' %}"
+    hx-include="[name='selected_incidents']"
+  {% else %}
+    method="post"
+    action="{{ endpoint|default:'htmx:incident-list' }}"
```
hmpf, line 9: This will not generate an url. Compare with line 5.

hmpf (changes requested):

src/argus_htmx/templates/htmx/incidents/_base_incident_update_modal.html
hmpf: Line 9, if endpoint is set, you will get the endpoint. If it is not set, you will get the string "'htmx:incident-list".

elfjes (changes requested):
I like where this is going. The htmx.trigger is a bit magic to me and some naming can be improved. But overall I like it

src/argus_htmx/incidents/views.py
```diff
@@ -110,6 +121,28 @@ def incident_detail_add_ack(request, pk: int, group: Optional[str] = None):
...
+        incident_ids = request.POST.getlist("selected_incidents", [])
+        form = formclass(formdata)
+        if form.is_valid():
+            cleaned_form = form.cleaned_data
+    return cleaned_form, incident_ids
+
+
+@require_POST
+def incidents_update(request: HtmxHttpRequest):
+    form_id = request.htmx.trigger_name or request.htmx.trigger
```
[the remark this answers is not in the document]
elfjes: Why use the form name and not just a hidden input (that would work for both htmx and non-htmx)?

src/argus_htmx/incidents/views.py
```diff
@@ -16,20 +17,30 @@
...
 DEFAULT_PAGE_SIZE = getattr(settings, "ARGUS_INCIDENTS_DEFAULT_PAGE_SIZE", 10)
 ALLOWED_PAGE_SIZES = getattr(settings, "ARGUS_INCIDENTS_PAGE_SIZES", [10, 20, 50, 100])
 
+# Map request trigger to parameters for incidents update
+INCIDENT_UPDATE_FORM_NAMES = {
+    "ack": (AckForm, bulk_ack_queryset),
+    "close": (DescriptionOptionalForm, bulk_close_queryset),
+    "reopen": (DescriptionOptionalForm, bulk_reopen_queryset),
+    "update-ticket": (EditTicketUrlForm, bulk_change_ticket_url_queryset),
+}
```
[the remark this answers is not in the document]
hmpf: I have a neater solution for it with classes but that can wait.

src/argus_htmx/incidents/views.py
```diff
@@ -110,6 +121,28 @@ def incident_detail_add_ack(request, pk: int, group: Optional[str] = None):
...
+        form = formclass(formdata)
+        if form.is_valid():
+            cleaned_form = form.cleaned_data
+    return cleaned_form, incident_ids
+
+
+@require_POST
+def incidents_update(request: HtmxHttpRequest):
+    form_id = request.htmx.trigger_name or request.htmx.trigger
+    formclass, queryset = INCIDENT_UPDATE_FORM_NAMES.get(form_id, (None, None))
```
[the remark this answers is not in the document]
hmpf: Eeek agreed

src/argus_htmx/incidents/views.py
```diff
@@ -135,11 +140,15 @@ def get_form_data(request, formclass: forms.Form):
 
 @require_POST
 def incidents_update(request: HtmxHttpRequest):
-    form_id = request.htmx.trigger_name or request.htmx.trigger
-    formclass, queryset = INCIDENT_UPDATE_FORM_NAMES.get(form_id, (None, None))
+    form_name = request.POST.get("form_name")
```
hmpf, line 143: This is still ugly.

elfjes: can't speak for @hmpf but I think the parameter name "form_name" is a bit confusing for a consumer (not a regular user, but someone who either programs against argus or looks at the devtools). They just want "do something" with the incidents. Whether or not there are forms associated with the implementation is irrelevant. Consider the following two options:
```
POST /incidents/update?form_name=close&selected_incidents=1&selected_incidents=2
```
vs
```
POST /incidents/update?action=close&incident_id=1&incident_id=2
```

which of the two looks more "natural"? I'd say the second one. Which is why I (still) recommend using the latter names instead of the former.

Btw, it is then totally ok to go:
```python
form_name  = request.POST.get("action")
incidents = request.POST.getlist("incident_id")
```

I concede that it's not a rest api, so this may not be as high a concern, but I think it still has value to make requests urls/params/bodies look nice and clean

src/argus_htmx/templates/htmx/incidents/_incident_acknowledge_modal.html
```diff
@@ -1,5 +1,6 @@
-{% extends "htmx/_base_form_modal.html" %}
+{% extends "htmx/incidents/_base_incident_update_modal.html" %}
 {% block dialogform %}
+  <input type="hidden" name="form_name" value="ack">
```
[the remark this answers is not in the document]
hmpf: Yeah

src/argus_htmx/incidents/views.py
hmpf: ```POST /incidents/update/close/?incident_id=1&incident_id=2```

is my preference.

With

```
def incidents_update(request: HtmxHttpRequest, action: str):
```

and 

```
path("update/<str:action>/", views.incidents_update, name="incidents-update"),
```

It's something other django developers will more intuitively understand.

hmpf: Not that we do it my way *now* but I'm extremely likely to rewrite it like that.

elfjes: that's a nice solution as well :)

hmpf (approved):

PR #167, opened 2024-10-10
Fix loading overlay

### LINKED ISSUE #165: Spinner overlay width and position need to be fixed (especially for small screen widths)
> Works for now. Overlay width and position need to be fixed (especially for small screen widths) but can be fixed in a separate PR:
Screenshot 2024-10-10 at 14 06 31

_Originally posted by @podliashanyk in https://github.com/Uninett/argus-htmx-frontend/pull/156#pullrequestreview-2360109183_

podliashanyk:
Closes #165 

Do we need to document the added CSS at all?


src/argus_htmx/tailwindtheme/snippets/20-loading-box.css
```diff
@@ -1,18 +1,25 @@
...
+      grid-column-start: 1;
+      grid-row-start: 1;
     }
-    .loading-box .loading-overlay:not(.htmx-request) {
-      display: none;
+    /* Adds overlay */
+    .loading-box:has(> .htmx-indicator.htmx-request) > *:not(.htmx-indicator) {
+      pointer-events: none;
+      opacity: 0.3;
     }
```
elfjes, line 24: ah, I guess `pointer-events: none;` make clicks "pass through" to the underlying element? That's nice. I had seen it before but never given it much thought :)

src/argus_htmx/templates/htmx/incidents/incident_list.html
```diff
@@ -10,12 +10,10 @@
             {% include "htmx/incidents/_incidents_menubar.html" %}
         {% endif %}
     </section>
-    <section id="incident-list" class="loading-box">
+    <section id="incident-list" class="loading-box overflow-x-auto">
```
elfjes, line 13: 🥳

src/argus_htmx/tailwindtheme/snippets/20-loading-box.css
```diff
@@ -1,18 +1,25 @@
 @layer components {
+   /*
+    * Example usage:
+    * <div class="loading-box">
+    *   <!-- Any content tag here -->
+    *   <div class="htmx-indicator">
+    *     <!-- Any optional feedback widget here -->
+    *   </div>
+    * </div>
+    */
```
elfjes, line 10: css and html looks cleaner so that's nice. It does mean that you can only ever put two element inside  `loading-box` but I guess that's fine :)

elfjes (approved):
Looks good. Minor comment

src/argus_htmx/tailwindtheme/snippets/20-loading-box.css
elfjes: fair point 👍

hmpf (approved):
Tested manually, looks good.

Bonus: the navbar keeps the same width as the table when the table is extra wide.
When the viewport is narrower than the table the navbar fits the viewport but that's fine.

PR #180, opened 2024-10-18
Add djLint to lint HTML

johannaengland:


pyproject.toml
```diff
@@ -41,3 +41,8 @@ output-format = "full"
 
 [tool.ruff.lint]
 select = ["E9", "F63", "F7", "F82"]
+ignore = ["E731"]
```
lunkwill42, line 44: What does this have to do with djLint?

.github/workflows/lint.yml
```diff
@@ -14,3 +14,20 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+
+    - name: Install dependencies
+      run: |
+        pip install -U pip
+        pip install djlint
```
lunkwill42, line 30: You will probably encounter the same issue with using pip to install tools globally once this workflow runs on an Ubuntu 24.04 worker in GitHub actions.  Perhaps consider the pipx route also here?

lunkwill42 (changes requested):
I was never part of the djLint discussions, but would like to take part in a discussion of what linter rules we would want to enable for this tool.

The PR looks mostly fine to me, just a couple of potential quirks...

pyproject.toml
johannaengland: It doesn't. This is a leftover from when it was based on a different version of main and I incorrectly rebased it, will fix this now

.github/workflows/lint.yml
johannaengland: I won't since one of the solutions to the global pip install is using the `Set up Python` step before (see: https://github.com/actions/runner-images/issues/10781#issuecomment-2414587300)

PR #181, opened 2024-10-18
Add djLint to format HTML

johannaengland:
This adds formatting HTML with djLint to pre-commit and Github actions. 

#182 actually formats the code.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -16,3 +16,20 @@ jobs:
       with:
         args: "format --check"
         changed-files: 'true'
+
+  djlint:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


lunkwill42:
This is marked as blocked for some (for me) unknown reason.  It seems to mostly overlap with #180, but I don't know why, @johannaengland

johannaengland:
> This is marked as blocked for some (for me) unknown reason. It seems to mostly overlap with #180, but I don't know why, @johannaengland

It is blocked since it cannot be merged before #182 is merged since this PR only adds the workflows etc without actually formatting the code.

johannaengland:
Merging since this was approved during live code review.

PR #182, opened 2024-10-18
Format codebase using djLint

johannaengland:
This is the PR where we can discuss if we want to use djLint to format our HTML templates. 

I used the following settings: 

```
[tool.djlint]
profile="django"
indent = 2
```

djLint documentation: https://djlint.com/


lunkwill42:
I have no issues with using a formatter also for templates, but I don't know what's been discussed by the team thus far (I was on vacation the week this was added).

johannaengland:
Merging now, since it was approved live.

PR #183, opened 2024-10-18
Format codebase using djade

johannaengland:
Now we can discuss if we prefer djade or djLint (#182) to format HTML files

Djade documentation: https://github.com/adamchainz/djade

### DIFF
--- a/src/argus_htmx/templates/htmx/incidents/_incident_row.html
+++ b/src/argus_htmx/templates/htmx/incidents/_incident_row.html
@@ -2,20 +2,20 @@
 	  HTMX-app
 	  {% block incident_pk %}
 	  {% include "htmx/incidents/_incident_pk.html" %}
-	  {% endblock %}
+	  {% endblock incident_pk %}
 	  {% block status %}
 	  {% include "htmx/incidents/_incident_status.html" %}
-	  {% endblock %}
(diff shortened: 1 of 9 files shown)


src/argus_htmx/templates/htmx/incidents/_incident_row.html
```diff
@@ -2,20 +2,20 @@
...
+	  {% endblock tag %}
 	  {% block start_time %}
 	  {% include "htmx/incidents/_incident_start_time.html" %}
-	  {% endblock %}
+	  {% endblock start_time %}
 	  {% block description %}
 	  {% include "htmx/incidents/_incident_description.html" %}
-	  {% endblock %}
+	  {% endblock description %}
   </li>
```
elfjes, line 21: I don't think incident_row.html is used anymore. We should get rid of (and also the unused ìncident_row` function in incidents/views.py)

elfjes (commented):
It doesn't seem to do much, does it?

johannaengland:
> It doesn't seem to do much, does it?

No, since it focuses on Django templates and not the underlying HTML

johannaengland:
We decided to go for djLint for formatting instead since we cannot use both at the same time (due to endblock rules on the same line) and djade alone is not enough.

PR #184, opened 2024-10-18
Use lowercase form method values

johannaengland:
This is to follow rule H029 of djLint.

Reference: https://www.djlint.com/docs/linter/#rules


elfjes (approved):
👍

PR #186, opened 2024-10-18
Merge bulk and singular incident update logic

### LINKED ISSUE #178: Use bulk endpoints when updating single incident
As suggested by @elfjes in https://github.com/Uninett/argus-htmx-frontend/pull/164#pullrequestreview-2356371687

podliashanyk:
Closes Uninett/argus-htmx-frontend#178. Blocked by Uninett/argus-htmx-frontend#164. Results in big cleanup.


src/argus_htmx/incidents/views.py
```diff
@@ -66,143 +67,37 @@ def incident_row(request, pk: int):
...
         form = formclass(formdata)
         if form.is_valid():
             cleaned_form = form.cleaned_data
-    return cleaned_form, incident_ids
+    return cleaned_form
 
 
 @require_POST
-def incidents_update(request: HtmxHttpRequest):
+def incidents_update(request: HtmxHttpRequest, pk: Optional[int] = None):
```
hmpf, line 88: Optional arguments to a view is a smell.

hmpf (commented):
What happened to autocreation of tickets? That is only relevant for single incidents.

src/argus_htmx/incidents/views.py
elfjes: If I may ask, why is that a smell? I quite like that pattern in general.

src/argus_htmx/incidents/views.py
```diff
@@ -55,119 +67,35 @@ def incident_row(request, pk: int):
...
-    return redirect("htmx:incident-detail", pk=pk)
+def incidents_update(request: HtmxHttpRequest, pk: Optional[int] = None):
+    form_id = request.htmx.trigger_name or request.htmx.trigger
+    formclass, queryset = INCIDENT_UPDATE_FORM_NAMES.get(form_id, (None, None))
+    formdata = get_form_data(request, formclass)
+    if formdata:
+        # If no list of selected incident pks is provided, assume singular incident update
+        incident_ids = request.POST.getlist("selected_incidents")
+        if not incident_ids:
+            incident_ids = [get_object_or_404(Incident, id=pk).pk]
```
hmpf, line 96: You would get rid of this double if with two functions or always using formdata.

hmpf (commented):
Remember to update to work with #207!

src/argus_htmx/incidents/urls.py
```diff
@@ -7,10 +7,6 @@
...
     path("", views.incident_list, name="incident-list"),
     path("<int:pk>/", views.incident_detail, name="incident-detail"),
-    path("<int:pk>/ack/", views.incident_add_ack, name="incident-add-ack"),
-    path("<int:pk>/ack-detail/", views.incident_detail_add_ack, name="incident-detail-add-ack"),
-    path("<int:pk>/close-detail/", views.incident_detail_close, name="incident-detail-close"),
-    path("<int:pk>/reopen-detail/", views.incident_detail_reopen, name="incident-detail-reopen"),
-    path("<int:pk>/ticket-edit-detail/", views.incident_detail_edit_ticket, name="incident-detail-edit-ticket"),
-    path("<int:pk>/ticket-add-detail/", views.incident_detail_add_ticket, name="incident-detail-add-ticket"),
+    path("update/", views.incidents_update, name="incidents-update"),
+    path("<int:pk>/update/", views.incidents_update, name="incidents-update"),
```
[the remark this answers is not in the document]
hmpf: (It tends to break the helper function "reverse".)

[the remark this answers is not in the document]
elfjes: I don't have a strong feelings for one way or the other, but the hx-vals approach seems nice and clean :)

elfjes (approved):

lunkwill42 (approved):
Not sure if the following issues really are related to this PR per se, but these are the issues I noticed while testing, all related to error handling:

Attempting to set a ticket URL as a bulk operation looks like this:

image

I'm told the URL needs to be a valid URL.  If I enter the above invalid URL, the page reloads and things look like they run just fine: No errors are displayed.  However, no ticket values are changed either.

If I enter a valid URL, the update-ticket endpoint returns a 500 error, with a traceback that suggest it failed because Argus was unable to communicate with Redis (which I hadn't started).  The UI, however, acts as if I never clicked the `Submit` button at all.


---

I was asked to test that this PR allows tickets to be *created* without "antics", but I'm not sure what the best approach to test ticket system integrations is in dev-mode, so I'd have to get some advice in that area first.

Overall, the changes look fine and worthwile, and they seem to work.  I would assume the abovementioned error handling issues are a systemic failing of argus-htmx-frontend at this early stage of development - but do we have issues registered to remember to fix them?

hmpf:
Bulk cannot do autocreation of tickets. It is not turned on for the details page either.

What I see when adding/changing a ticket on the details page is that after opening the popup, entering a valid url and clicking the submit button in the popup, nothing happens. If I close the popup and reload the page, there's suddenly a ticket url in the ticket field.

hmpf:
When testing on current master and main I can add/change tickets just fine, and there are change-events made, auditing the change of the ticket_url field.

PR #194, opened 2024-10-22
Remove unused imports in Python code and unused dependency

johannaengland:
To add F4 rules to ruff. 

Reference: https://docs.astral.sh/ruff/rules/#pyflakes-f

Also removes a dependency that had become unused.

### DIFF
--- a/src/argus_htmx/models.py
+++ b/src/argus_htmx/models.py
@@ -1,3 +1,3 @@
-from django.db import models
+from django.db import models  # noqa: F401 - unused-import
 
 # Create your models here.
--- a/src/argus_htmx/tests.py
+++ b/src/argus_htmx/tests.py
@@ -1,3 +1,0 @@
-from django.test import TestCase
-
-# Create your tests here.
--- a/src/argus_htmx/themes/views.py
+++ b/src/argus_htmx/themes/views.py
@@ -1,13 +1,7 @@
-import importlib_resources
 import logging
-from pathlib import Path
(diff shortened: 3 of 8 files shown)


lunkwill42 (commented):
I think this is a worthwhile linter rule.  Unused imports tend to clutter the codebase as it grows unless we have something to remind us to prune them.  @hmpf ?

src/argus_htmx/tests.py
```diff
@@ -1,3 +1,3 @@
-from django.test import TestCase
+from django.test import TestCase  # noqa: F401 - unused-import
```
hmpf, line 1: Better to remove the file, it's autogenerated.

src/argus_htmx/models.py
```diff
@@ -1,3 +1,3 @@
-from django.db import models
+from django.db import models  # noqa: F401 - unused-import
```
hmpf, line 1: This file will be used by the userprefs-branch and should stay even if it was autogenerated.

src/argus_htmx/themes/views.py
```diff
@@ -1,13 +1,7 @@
-import importlib_resources
```
hmpf, line 1: This has a dependency that can be removed, see "pyproject.toml".

hmpf (changes requested):
This renders a dependency redundant, see comment.

hmpf (approved):

PR #195, opened 2024-10-22
Use old percentage syntax for log messages

johannaengland:


lunkwill42 (commented):
I would want this :)

pyproject.toml
```diff
@@ -40,4 +40,4 @@ line-length = 120
 output-format = "full"
 
 [tool.ruff.lint]
-select = ["E9", "F63", "F7", "F82"]
+select = ["E9", "F5", "F63", "F7", "F82"]
```
hmpf, line 43: Yes

src/argus_htmx/incidents/views.py
```diff
@@ -123,7 +123,7 @@ def incident_detail_close(request, pk: int):
             request.user,
             description=form.cleaned_data.get("description", ""),
         )
-        LOG.info(f"{{ incident }} manually closed by {{ request.user }}")
+        LOG.info(f"{ incident } manually closed by { request.user }")
```
hmpf, line 126: This should use the old percent syntax: `Log.info("str with %", arg1, arg2)`

hmpf: Same with the other log-message.

johannaengland:
Merging since this was discussed during a live code review.

PR #196, opened 2024-10-22
Add F5, F6, F8 and F9 rules to ruff

johannaengland:
They are already fulfilled on the current codebase. 

Reference: https://docs.astral.sh/ruff/rules/#pyflakes-f


lunkwill42 (commented):
I'm all for it :)

johannaengland:
Merging since this was approved in a live code review.

PR #197, opened 2024-10-22
Add E4 rules to ruff

johannaengland:
Codebase already fulfills these rules. 

Reference: https://docs.astral.sh/ruff/rules/#error-e

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ line-length = 120
 output-format = "full"
 
 [tool.ruff.lint]
-select = ["E7", "E9", "F5", "F6", "F7", "F8", "F9"]
+select = ["E4", "E7", "E9", "F5", "F6", "F7", "F8", "F9"]
 
 [tool.djlint]
 profile="django"


lunkwill42 (commented):
Once we get into this class of linter violations: Won't they already be flagged by attempting to run black or ruff formatting of the codebase?

johannaengland:
> Once we get into this class of linter violations: Won't they already be flagged by attempting to run black or ruff formatting of the codebase?

I tested this by adding `import os, sys` and then running `ruff format .` and it was not changed.

lunkwill42:
> I tested this by adding `import os, sys` and then running `ruff format .` and it was not changed.

My bad. I tried to google what the `E4` group of violations mean, and was left with the impression that they were all kinds of syntax errors: But now I'm thinking I must have found a reference from an entirely different tool, because it does not match up at all with the ruff docs we just looked.

Please disregard my earlier comment, I'm all for this.

lunkwill42 (approved):

PR #198, opened 2024-10-22
Rewrite assigned lambda expressions as def

johannaengland:
To add E7 rules to ruff.

Reference: https://docs.astral.sh/ruff/rules/#error-e


lunkwill42 (approved):
Yes. That is all.

PR #202, opened 2024-10-24
Add and apply TailwindCSS Typography plugin

podliashanyk:
Relevant for Uninett/Argus#1054

#### Discussion:

- should typography plugin be activated (current implementation) or deactivated on the `<main>` by default?

### DIFF
--- a/src/argus_htmx/static/styles.css
+++ b/src/argus_htmx/static/styles.css
@@ -853,6 +853,511 @@ html {
   --tw-contain-style:  ;
 }
 
+.prose {
+  color: var(--tw-prose-body);
+  max-width: 65ch;
+}
+
+.prose :where(p):not(:where([class~="not-prose"],[class~="not-prose"] *)) {
+  margin-top: 1.25em;
+  margin-bottom: 1.25em;
+}
+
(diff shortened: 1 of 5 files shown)


elfjes:
What's this about?

elfjes (changes requested):
My initial/gut reaction about using @tailwindcss/typography: Let's be really careful before adding this.

Reading the docs, it seems that the `prose` class is meant for larger sections that have multiple headings, paragraphs, etc (such as content coming from a CMS or rendered markdown). It's called "prose" for a reason. The docs explicitly say that you probably don't want default prose styling, and I agree. The result is that now you need to not-prose large sections of the page, which seems like an antipattern.

If we want to apply some styles to headings because we want a heading system, that is fine, but I'd suggest creating classes for that and not just add style to tags by default.

lunkwill42 (commented):
I read your comments about the typography plugin in my PR, but I'm not sure I understand the reasoning, given how the plugin is explained.

We're not about to add large sections of prose to Argus, so I find this a bit overkill just to style some page headings.

hmpf:
should our styles have a prefix? "argus-"?

hmpf:
I'm closing this...

PR #205, opened 2024-10-25
Add destinations page

### LINKED ISSUE #169: Make destinations-page

stveit:
Fixes Uninett/Argus#1001 

Replaced by Uninett/Argus#970

Adds page where you can see and edit existing destinations and create new ones.

Since we are kinda locked to using the serializer for validating destinations (for now, changes will be made at some point) I ended up baking the serializer into the forms for creating/updating destinations. Its a bit wonky but should work fine until we can move on from the serializer and have `ModelForm` do everything they way it is supposed to.

Needs more styling to be complete, but not sure how to do it properly.. Error in fields are now added to the right of the field the error belongs to, so the buttons and everything gets moved to the right when theres an error.


hmpf:
Lookswise:

Having "Label", "Media", "Value" inside the same box as the actual changeable value looks terrible. I think the login form has the value on the outside.

There should be the rounded box around each if we want to have boxes, again see existing code.

I'm not sure the dropdowns are needed at all. We should not expect a zillion destinations per user.

The grey behind each box has to go. Too low contrast.

The "media" input in each box is redundant.

I think a header would look nicer than having one box for each type.

hmpf:
Functions wise: make it work without HTMx first. Django has something called "formsets" you can use, see the abandoned code in #109 (which tries to do too much at once as well).

hmpf:
DELETE and PATCH is API-stuff and htmx-stuff, not plain HTML. Make it work with plain HTML first.

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,9 @@
+from django import forms
+
+from argus.notificationprofile.models import Media
+
+
+class DestinationForm(forms.Form):
```
hmpf, line 6: Could probably be a ModelForm.

src/argus_htmx/destinations/urls.py
```diff
@@ -1,21 +1,11 @@
...
-
+from . import views
 
 app_name = "htmx"
 urlpatterns = [
-    path("", placeholder, name="destination-placeholder"),
+    path("", views.destinations, name="destinations"),
+    path("create_destination/", views.destinations_create, name="destinations-create"),
+    path("update_destination/<int:pk>/", views.destinations_update, name="destinations-update"),
+    path("delete_destination/<int:pk>/", views.destinations_delete, name="destinations-delete"),
```
hmpf, line 10: POST to /destination/: create

POST to /destination/<int:pk>/: update

POST to /destination/<int:pk>/: delete

This is the old, well tested way.

src/argus_htmx/destinations/views.py
```diff
@@ -0,0 +1,81 @@
...
+            initial={
+                "label": destination.label,
+                "media": destination.media.slug,
+                "value": destination.settings.get("email_address", ""),
+            }
+        )
+        form.fields["media"].widget.attrs["hidden"] = True
+        grouped_destinations[destination.media].append((destination, form))
+
+    return grouped_destinations
```
hmpf, line 46: Not needed with ModelForm. You still might want to group though: sort by type, then label.

hmpf (commented):
I think this is a good candidate for class based views actually. We should have a meeting.

src/argus_htmx/destinations/urls.py
stveit: Made it like this, so now /destination/ is used to get the site itself with GET and POST to create a destination

src/argus_htmx/destinations/views.py
```diff
@@ -1,81 +1,128 @@
...
+                "media": media,
+                "label": form.cleaned_data.get("label", ""),
+                "settings": {settings_key: form.cleaned_data["value"]},
+            },
+            context={"request": request},
         )
-        destination.save()
+        serializer.is_valid(raise_exception=True)
+        serializer.save(user=request.user)
+
```
hmpf, line 76: I was thinking you'd use the form to save, not go via a serializer. There's magic needed but we all ought to know that magic.

src/argus_htmx/destinations/views.py
```diff
@@ -1,81 +1,128 @@
...
 @require_POST
 def destinations_update(request, pk: int) -> HttpResponse:
-    form = DestinationForm(request.POST or None)
+    form = DestinationFormUpdate(request.POST or None)
     if form.is_valid():
         destination = DestinationConfig.objects.get(pk=pk)
-        destination.label = form.cleaned_data.get("label", "")
-        destination.settings = {"email_address": form.cleaned_data["value"]}
-        destination.save()
+        settings_key = _get_settings_key_for_media(destination.media)
```
hmpf, line 85: You don't need line 84.

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,79 @@
...
+            # Serializer might add more data to the JSON dict
+            if settings := self.serializer.validated_data.get("settings"):
+                self.cleaned_data["settings"] = settings
+
+        if label := self.cleaned_data["label"]:
+            filter = DestinationConfig.objects.filter(label=label)
+            if self.instance:
+                filter = filter.exclude(pk=self.instance.pk)
+            if filter.exists():
+                self.add_error("label", "Name must be unique per media")
```
stveit, line 69: If label is `""` then this wont work

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,79 @@
...
+        model = DestinationConfig
+        fields = ["label", "media", "settings"]
+        labels = {
+            "label": "Name",
+        }
+
+    def clean(self):
+        super().clean()
+        settings_key = _get_settings_key_for_media(self.cleaned_data["media"])
+        self.cleaned_data["settings"] = {settings_key: self.cleaned_data["settings"]}
```
stveit, line 26: Add comment to explain that this is for making the value compatible with JSONField, which is what DestinationConfig.settings should really be. The form overrides it to be a CharField for the frotnend to make more sense (we shouldnt expect the user to give a full valid dict in JSON format as input, only the email address / phone number itself should be requires)

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,79 @@
...
+                    if error_name == settings_key:
+                        error_name = "settings"
+                    self.add_error(error_name, error_detail)
+                    # Serializer might add more data to the JSON dict
+                    if settings := self.serializer.data.get("settings"):
+                        self.cleaned_data["settings"] = settings
+        else:
+            # Serializer might add more data to the JSON dict
+            if settings := self.serializer.validated_data.get("settings"):
+                self.cleaned_data["settings"] = settings
```
stveit, line 62: should maybe flip on the logic here, have valid state, first, and not valid under else

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,79 @@
...
+from argus.notificationprofile.models import DestinationConfig, Media
+from argus.notificationprofile.serializers import RequestDestinationConfigSerializer
+from argus.notificationprofile.media import api_safely_get_medium_object
+
+
+class DestinationFormCreate(ModelForm):
+    settings = forms.CharField(required=True)
+
+    def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request", None)
```
stveit, line 13: Add comment stating this is needed because the serializer needs access to the request object

src/argus_htmx/destinations/forms.py
```diff
@@ -71,6 +71,47 @@ def _validate_serializer(self):
         return self.cleaned_data
 
 
+class DestinationFormUpdate(DestinationFormCreate):
+    def __init__(self, *args, **kwargs):
+        if instance := kwargs.get("instance"):
+            settings_key = _get_settings_key_for_media(instance.media)
+            instance.settings = instance.settings.get(settings_key)
+        super().__init__(*args, **kwargs)
```
stveit, line 79: add comment explaining that this is done to convert the settings value from the whole dict to just the email/phone number. That way we dont get the entire dict in the frontend, and just the value we are interested in. This is opposite to what is done in the create form, where we translate from the value itself to a whole dict

src/argus_htmx/destinations/forms.py
stveit: But it might be fine? `""` seems to get translated to None when the object is saved to the database, and the constraints seem to allow multiple destinations without a label.

But the thing that should be changes is `self.cleaned_data["label"] `should be `self.cleaned_data.get("label")`. This doesnt cause any issues when testing, because the frontend sends an empty string for label if the input box is empty, but since this form doesnt actually require label to be set, we should assume that there are situations where it matters.
```suggestion
        if label := self.cleaned_data.get("label"):
            filter = DestinationConfig.objects.filter(label=label)
            if self.instance:
                filter = filter.exclude(pk=self.instance.pk)
            if filter.exists():
                self.add_error("label", "Name must be unique per media")
```

stveit: Additionally, it might be wrong to check the uniqueness of the name here in the form validator. There is already a constraint in the DestinationConfig model for exactly this. This could then be handled by trying to save and see if the constraint throws an exception and take it from there. This is more or less a repetition of that constraint. Buut you would need some code anyway, and it might be easier to just let the form handle everything, instead of having to catch the constraint exception and cram errors into the form from the outside

src/argus_htmx/destinations/views.py
```diff
@@ -0,0 +1,106 @@
...
+        if update_form.instance.pk == pk:
+            update_forms[index] = form
+            break
+    return _render_destinations(request, update_forms=update_forms)
+
+
+def _render_destinations(
+    request,
+    create_form: Optional[DestinationFormCreate] = None,
+    update_forms: list[DestinationFormUpdate] = None,
```
stveit, line 65:
```suggestion
    update_forms: Sequence[DestinationFormUpdate] = None,
```

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,120 @@
...
+
+        self.serializer = RequestDestinationConfigSerializer(
+            destination,
+            data=data,
+            context={"request": self.request},
+            partial=True,
+        )
+
+
+def _get_settings_key_for_media(media: Media) -> str:
```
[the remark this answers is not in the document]
hmpf: As is, assumes that there is only one setting a user can set per plugin, which currently holds but might not in the future.

src/argus_htmx/destinations/forms.py
```diff
@@ -0,0 +1,120 @@
...
+            # Serializer might add more data to the JSON dict
+            if settings := self.serializer.validated_data.get("settings"):
+                self.cleaned_data["settings"] = settings
+
+        if label := self.cleaned_data["label"]:
+            filter = DestinationConfig.objects.filter(label=label)
+            if self.instance:
+                filter = filter.exclude(pk=self.instance.pk)
+            if filter.exists():
+                self.add_error("label", "Name must be unique per media")
```
hmpf, line 69: Hopefully all of this will go away when the plugins are changed.

src/argus_htmx/templates/htmx/destinations/destinations.html
hmpf, line 1: Should be named destination/destination_list.html

src/argus_htmx/templates/htmx/destinations/edit_or_delete_form.html
hmpf, line 1: Should be named "destination/_edit_or_delete_form.html"

hmpf: .. and split it into three files. On to hold em, and one per action:

- destination/_forms.html
- destination/_edit_form.html
- destination/_delete_form.html

src/argus_htmx/destinations/forms.py
hmpf, line 1: Rename "argus_htmx/destination*_s_*/" to the singular: "argus_htmx/destination/"

hmpf (changes requested):
I've tested it and it works except it's a bit confusing that when I edit/delete the url changes. Maybe redirect to "/destination/" after change? @podliashanyk might have a better idea.

Almost only renames following.

hmpf:
Ehh, and I almost forgot: media-field of the creation-form is bugged in "argus"-theme:
image

src/argus_htmx/destinations/forms.py
stveit: :crossed_fingers:

stveit:
> I've tested it and it works except it's a bit confusing that when I edit/delete the url changes. Maybe redirect to "/destination/" after change? @podliashanyk might have a better idea.
> 
> Almost only renames following.

I imagine HTMXifying this properly should fix this. Instead of loading the entire page when updating, only the update form you used is actually updated etc

hmpf:
Moved to argus-server

PR #212, opened 2024-10-29
Create a place to store user preferences in the database for this app

hmpf:
Created in tandem with and dependent on Uninett/Argus#889


src/argus_htmx/themes/views.py
```diff
@@ -21,33 +21,43 @@ class ThemeListView(ListView):
...
 
 @require_POST
 def change_theme(request: HtmxHttpRequest) -> HttpResponse:
-    themes = get_theme_names()
-    theme = request.POST.get("theme")
-    if theme in themes:
-        request.session["theme"] = theme
-        messages.success(request, f'Switched theme to "{theme}"')
-        return HttpResponse(f"{theme}")
+    prefs = request.user.get_preferences("argus_htmx")
```
elfjes, line 49: What about unauthenticated users?

src/argus_htmx/context_processors.py
```diff
@@ -7,28 +7,12 @@
...
-    datetime_format = DATETIME_FORMATS[datetime_format_name]
-    return {
-        "datetime_format": datetime_format,
-        "datetime_format_name": datetime_format_name,
-    }
+def preferences(request):
+    pref_sets = Preferences.objects.filter(user=request.user)
+    prefdict = {}
+    for pref_set in pref_sets:
+        prefdict[pref_set._namespace] = pref_set.get_context()
```
elfjes, line 17: This makes sense

src/argus_htmx/dateformat/views.py
```diff
@@ -21,10 +21,10 @@ def dateformat_names(request: HtmxHttpRequest) -> HttpResponse:
 
 @require_POST
 def change_dateformat(request: HtmxHttpRequest) -> HttpResponse:
-    datetime_format_name = request.POST.get("dateformat", DATETIME_DEFAULT)
-    default_format = DATETIME_FORMATS[DATETIME_DEFAULT]
-    datetime_format = DATETIME_FORMATS.get(datetime_format_name, default_format)
-    request.session["datetime_format"] = datetime_format
-    request.session["datetime_format_name"] = datetime_format_name
-    messages.success(request, f'Switched dateformat to "{datetime_format_name}"')
+    prefs = request.user.get_preferences("argus_htmx")
```
elfjes, line 24: Maybe we can prefetch (select_related) all a user's preferences?

src/argus_htmx/themes/views.py
```diff
@@ -21,33 +21,43 @@ class ThemeListView(ListView):
...
 
     def post(self, request, *args, **kwargs):
-        theme = request.POST.get("theme")
-        if theme in self.themes:
-            request.session["theme"] = theme
+        prefs = request.user.get_preferences("argus_htmx")
+        form = prefs.FORMS["theme"](request.POST)
+        if form.is_valid():
+            theme = form.cleaned_data["theme"]
+            prefs.save_preference("theme", theme)
```
elfjes, line 34: This is a common pattern, can we extract/abstract it out?

src/argus_htmx/user/preferences/models.py
```diff
@@ -0,0 +1,58 @@
...
+    def create(self, **kwargs):
+        kwargs["namespace"] = ArgusHtmxPreferences._namespace
+        return super().create(**kwargs)
+
+
+class ArgusHtmxPreferences(Preferences):
+    _namespace = "argus_htmx"
+    FORMS = {
+        "datetime_format_name": DateTimeFormatForm,
+        "page_size": PageSizeForm,
```
elfjes, line 46: rename to `incident_list_page_size` or something akin. there will probably be other listings

hmpf: I hope not actually. There ought not be so many destinations, profiles, timeslots etc. per user that paging (or a table) is needed.

src/argus_htmx/themes/views.py
hmpf: We'd probably have to copy preferences to the session. The session (depending on how it is stored) has a max. size though.

src/argus_htmx/user/preferences/models.py
elfjes: still, can't hurt to be explicit about its purpose

hmpf: Worst case: even more settings...

src/argus_htmx/themes/views.py
hmpf: I'd love to.

src/argus_htmx/dateformat/views.py
hmpf: Where? When?

elfjes: when fetching the request's user. I was thinking it saves us some db round trips ,but perhaps it's not the effort. We currently have one or two queries for preferences, one in the context processor, and perhaps one in the view. So any gains are probably minor

src/argus_htmx/themes/views.py
elfjes: The session only needs to store preferences that are different from the default, so that's probably ok. Also, I'm not quite sure how big the max size is?

hmpf: Session storage in a cookie is limited by max cookie size. Sessions can also be stored in for instance the database, with just the key in a cookie, and then there are no limits.

src/argus_htmx/dateformat/views.py
hmpf: request.user is fetched in the auth middleware IIRC.

elfjes: right, that would probably make it too tricky, for only a minor gain

src/argus_htmx/themes/views.py
elfjes: As far as I can see, sessions are stored in db by default (using the contrib.sessions app)

hmpf: Right now it is stored in session iff the user is not authenticated.

elfjes: what setting governs that?

elfjes (approved):

PR #213, opened 2024-10-30
Source selector with dropdown

podliashanyk:
Closes Uninett/argus-htmx-frontend#103 

### Implemented changes:

- added reusable generic templates for rendering multi-select with a dropdown
- implemented sources selector in incidents filter as multi-select with dropdown

### Further work:
- Uninett/Argus#1061 
- polish table transitions on filter updates Uninett/Argus#1064 

### Changes in UI:

- Old sources selector:
Screenshot 2024-10-30 at 11 12 58

- New sources selector with selected elements displayed as badges (current implementation for sources):
Screenshot 2024-10-30 at 11 11 42

- Default rendering of selection results in the generic template (using sources selector just as an example here):
Screenshot 2024-10-30 at 11 12 09

### DIFF
--- a/src/argus_htmx/templates/htmx/forms/checkbox_select_multiple.html
+++ b/src/argus_htmx/templates/htmx/forms/checkbox_select_multiple.html
@@ -1,9 +1,12 @@
-{% if widget.wrap_label %}<label{% if widget.attrs.id %}
-  for="{{ widget.attrs.id }}"{% endif %}
-  class="flex gap-2 items-center my-1"
->{% endif %}
-<input type="{{ widget.type }}" name="{{ widget.name }}"{% if widget.value != None %}
-       class="checkbox checkbox-sm checkbox-primary border"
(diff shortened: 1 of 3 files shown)


hmpf:
This will be useful for the profiles-page but I'm guessing it is not sufficient for tag-input?

src/argus_htmx/incidents/filter.py
```diff
@@ -9,14 +9,21 @@
 QuerySetFilter = filter_backend.QuerySetFilter
 
 
+class DropdownMultiSelect(forms.CheckboxSelectMultiple):
+    template_name = "htmx/incidents/_incident_source_select.html"
+    option_template_name = "htmx/forms/checkbox_select_multiple.html"
```
hmpf, line 14: Older Django didn't have this! Nice, huh?

hmpf (approved):
I hope that when there is a dropdown icon, the placeholder text will disappear once anything has been selected since it makes it so wide.

For search you need a typeahead endpoint first I guess.

PR #223, opened 2024-11-04
Integrate with django messages to display messages as toasts

elfjes:


src/argus_htmx/settings.py
```diff
@@ -56,3 +57,10 @@
 DAISYUI_THEMES = get_json_env("DAISYUI_THEMES", DEFAULT_THEMES, quiet=True)
 DEFAULT_THEME_OVERRIDE = {}
 TAILWIND_THEME_OVERRIDE = get_json_env("TAILWIND_THEME_OVERRIDE", DEFAULT_THEME_OVERRIDE, quiet=True)
+
+MESSAGE_TAGS = {
+    messages.INFO: "alert-info",
+    messages.SUCCESS: "alert-success",
+    messages.WARNING: "alert-warning",
+    messages.ERROR: "alert-error",
+}
```
hmpf, line 66: Have you checked that this doesn't break the admin?

src/argus_htmx/templates/tailwind/tailwind.config.js
```diff
@@ -16,7 +16,8 @@ module.exports = {
         ...{{ themeoverride }},
     },
     safelist: [
-        "htmx-request"
+        "htmx-request",
+        { pattern: /alert-(info|success|warning|error)/ }
```
hmpf, line 20: Add in the PR description that the tailwind config needs to be rebuilt.

src/argus_htmx/settings.py
elfjes: good point

src/argus_htmx/templates/tailwind/tailwind.config.js
elfjes: yes. Ideally we'd pick up on these classes automatically but that requires specifying perhaps every `.py` file in argus_htmx as possible css class source (project dir) (and then also publishing that so that tailwind_config can automatically pick up on that, but that requires some additional discoverability

src/argus_htmx/settings.py
elfjes: django admin seems fine, but could you perhaps verify?

elfjes: decided to stick with default tags

hmpf:
Why does the login-page have a different system to show messages?

elfjes:
> Why does the login-page have a different system to show messages?

I guess that's a matter of preference. To me it seems more natural to get those messages above the login box, than as a toast at the right bottom corner. It's the common ux. And since we don't have any htmx/async requests on the login page, that's ok. On the incidents page we do have asynchronous requests, so there it makes more sense to put them as toasts

src/argus_htmx/templates/messages/_notification_message.html
```diff
@@ -0,0 +1,17 @@
+<div class="alert flex justify-between w-fit border-3 bg-base-100 border-opacity-100 {% if message.tags %}{{ message.tags }}{% endif %}">
```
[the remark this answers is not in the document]
elfjes: This is a django messages feature. every messages gets assigned some tags: the message level, and any additional tags given to the message when creating it. these tags are space-separated string-concatenated to the message object and meant to be used as html classes. see https://docs.djangoproject.com/en/5.1/ref/contrib/messages/#message-tags

src/argus_htmx/tailwindtheme/snippets/15-alert-message.css
```diff
@@ -0,0 +1,23 @@
+/* use color-mix to give the illusion of transparency while not actually being transparent */
+@layer components {
+  .alert.info {
+    @apply border-info;
+    background-color: color-mix(in srgb, oklch(var(--in)), oklch(var(--b1)) 85%);
+
```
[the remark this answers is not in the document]
elfjes: thanks :)

elfjes:
> > > Why does the login-page have a different system to show messages?
> > 
> > 
> > I guess that's a matter of preference. To me it seems more natural to get those messages above the login box, than as a toast at the right bottom corner. It's the common ux. And since we don't have any htmx/async requests on the login page, that's ok. On the incidents page we do have asynchronous requests, so there it makes more sense to put them as toasts
> 
> Hmm, I would rather put them directly below the login box or at the very least ensure overflow behaviour for messages box. We don't want the login box to disappear from the view if there are many messages on login page.

Hmm since we're sending out messages on every response, the chance of having many messages appear at once is limited I'd say. Still, can't hurt to be cautious :)

elfjes:
> Good stuff! See some comments and questions below.
> 
> A polish that I consider a useful UX would be to have some toasts to auto-close (see [remove-me HTMX extension](https://github.com/bigskysoftware/htmx-extensions/blob/main/src/remove-me/README.md). "Switched theme..." toast is a good candidate where I would rather be informed about it without having to close the alerts manually. A task for a separate PR though.

Yeah that would be nice, might be something we achieve by reading the tags. A success or info tag gets auto removed, while an error tag stays. Or maybe if a certain custom tag exists. The hyperscript would be something like `_="wait 5s then remove me"`

> 
> Also we will need to add overflow behaviour at some point (both in Y and X plane). Can be a separate PR

yeah good point, or we could start a second column once the first one is full :D. But with autoclosing messages there is less opportunity for overflow

> 
> Also I would prefer to be able to close error toasts that were caused by some form in a modal without closing the modal. Per now all toasts are located behind the modal overlay. It is out of scope of this PR though. It might also happen that this is a non-issue and that we might end up on showing **custom** toasts for modal errors (something like on login page).

hmm yeah, I hadn't looked at modals yet. perhaps just a matter of setting the z-index

elfjes:
> I'm fond of the standard Django way to do it which is always at the top, before the header, but then I am old-fashioned.
> 
> Failed user/password logins should probably be handled in the form itself, no messages needed.

I noticed that django_social hooks into the messages framework, which was what got me in this track in the first place. I think we should strive to have a single place where we put login errors (apart from validation errors)

elfjes:
> Hmm, I would rather put them directly below the login box or at the very least ensure overflow behaviour for messages box. We don't want the login box to disappear from the view if there are many messages on login page.

Hmm I've put them below the login box now, but it looks a little out of place, but maybe that's just me

image

elfjes:
> > Also I would prefer to be able to close error toasts that were caused by some form in a modal without closing the modal. Per now all toasts are located behind the modal overlay. It is out of scope of this PR though. It might also happen that this is a non-issue and that we might end up on showing **custom** toasts for modal errors (something like on login page).
> 
> hmm yeah, I hadn't looked at modals yet. perhaps just a matter of setting the z-index

Hmm there are some css shenanigans afoot but I'm having trouble with rendering the notifications above the/a modal. Something for another PR then. If we go the custom-message way, we'd have to make sure to have a good common base template for the modal (we have modals in a number of different places, not just the ones for doing an action on an incident, and it would be nice to have a consistent ux wrt modals)

elfjes:
> I prefer it this way 😄 Also reduces the acute need for transitions on login box since the pop-up toast is located below the "static" login box so there is less netto moving around of the elements

Alright, let's keep it like this then :)

elfjes:
> I would say a vertical scroll if height threshold for the `#notification-messages` container is reached. The threshold could be either a viewport, or even better ensuring that navbar is never covered by toasts. This constraints work with or without [daisyUI stack](https://daisyui.com/components/stack/).
> 
> For the horizontal overflow I would prefer to add the constrains to each edit:~toast~ alert itself vs to the `#notification-messages` container. We could go for a more nuanced behaviour which combines appropriate mix of "word break+width+max-width+horizontal grow+max-height+vertical grow. But I suggest we instead start with a simple horizontal scroll i x plane for each individual ~toast~ alert if max viewport width is reached.
> 
> What do you think?



I'd think a vertical scroll-bar on the `#notification-messages` may look strange. I think it would not be obivous what that scrollbar is for.  I'm actually not all sure we need diferent vertical overflow behaviour. Currently it pushes old messages outside the viewport, which will reappear once you start closing (newer) messages. I don't necessarily see unwanted behaviour their. There may be a performance issue if all those messages are being rendered but not shown, but I'm not sure if that is any different with different overflow behaviour. We need to make sure they're rended on top of the user menu, but other than that?

horizontal overflow behaviour: I'd say we nee to set a maximum width (possbily responsive) and then word wrap the contents. Perhaps a maximum height for the notification as well? Which we could either ensure using a scroll or ellipsis

PR #226, opened 2024-11-06
Wash list of theme names

hmpf:
Ensure that the themes shown in the theme-selector are actually installed.


src/argus_htmx/themes/utils.py
```diff
@@ -1,17 +1,42 @@
...
+    for theme in themes_setting:
         if isinstance(theme, str):
             theme_names.append(theme)
         elif isinstance(theme, dict):
             theme_names.extend(theme.keys())
     return theme_names
+
+
+def get_themes_from_css():
+    styles_css = files("argus_htmx").joinpath("static/styles.css").read_text()
```
elfjes, line 24: The css file is overridable using the `path_to_stylesheet` context variable, so this should respect that

src/argus_htmx/themes/utils.py
```diff
@@ -1,17 +1,42 @@
...
+    theme_names = []
+    for line in styles_css:
+        if "data-theme_name=" not in line:
+            continue
```
elfjes, line 28: css file may be minified and have everything on one line

elfjes (changes requested):

src/argus_htmx/themes/utils.py
hmpf: This file is the one delivered with the code. Not much to gain in minifing it. The ones used by the servers (after collectstatic) is another matter but our python code should not assume that code is accessible in any way.

src/argus_htmx/themes/utils.py
hmpf: Template context can never be available at this stage in the loading of the app.

We'd need to change how `path_to_stylesheet` is set, if not `path_to_stylesheet` is also only valid after `collectstatic` is run.

src/argus_htmx/themes/utils.py
elfjes: I'm not suggesting to minify the current default argus styles.css, I'm saying that we have a minified styles.css running in deployment, at non-default location (see also my comment about path_to_stylesheet). We would get errors if we try to start argus in deployment.

src/argus_htmx/themes/utils.py
elfjes: I would be ok doing that through a setting

hmpf: Setting made, mentioned in readme.

src/argus_htmx/settings.py
```diff
@@ -57,3 +58,4 @@
 THEME_DEFAULT = get_str_env("ARGUS_THEME_DEFAULT", "argus")
 DEFAULT_THEME_OVERRIDE = {}
 TAILWIND_THEME_OVERRIDE = get_json_env("TAILWIND_THEME_OVERRIDE", DEFAULT_THEME_OVERRIDE, quiet=True)
+STYLESHEET_PATH = get_str_env("STYLESHEET_PATH", STYLESHEET_PATH_DEFAULT)
```
elfjes, line 61: should the env be ARGUS_STYLESHEET_PATH?

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,50 @@
...
+    stylesheet_path = static_url / default_htmx_settings.STYLESHEET_PATH
+    styles_css = files("argus_htmx").joinpath(stylesheet_path).read_text()
+    styles_css_lines = styles_css.split("{")
+    theme_names = []
+    for line in styles_css_lines:
+        if "data-theme=" not in line:
+            continue
+        _, after = line.split("=", 1)
+        theme_name, _ = after.split("]", 1)
+        theme_names.append(theme_name.strip())
```
elfjes, line 35: wouldn't it be cleaner and less fragile to use an `re.findall` search here?

hmpf: Easier to read for others, probably yes. I like to avoid regexes when I can.

elfjes: which is a good sentiment, but sometimes they make life easier and can actually make sense ;)

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,47 @@
...
             theme_names.extend(theme.keys())
     return theme_names
 
 
+def get_themes_from_css():
+    THEME_NAME_RE = "(?P<theme>\w+)"
+    DATA_THEME_RE = f"\[data-theme={THEME_NAME_RE}\]"
+
+    static_url = Path(settings.STATIC_URL).relative_to("/")
+    stylesheet_path = static_url / default_htmx_settings.STYLESHEET_PATH
```
elfjes, line 30: this needs to look at the django.conf.settings since it might  be overridden in another settings file (we do this)

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,47 @@
...
+    installed_themes = themes_from_setting | themes_from_css
+    all_themes = themes_from_setting & themes_from_css
+    if all_themes != installed_themes:
+        raise ImproperlyConfigured("Themes in settings is out of sync with themes installed")
+    return installed_themes
+
+
 def get_theme_default():
-    return getattr(settings, "THEME_DEFAULT", argus_htmx_settings.THEME_DEFAULT)
+    return getattr(settings, "THEME_DEFAULT", default_htmx_settings.THEME_DEFAULT)
```
elfjes, line 47: this should also look at django.conf.settings

hmpf: It already does. `default_htmx_settings.THEME_DEFAULT` is the fallback.

hmpf: I have renamed "default_htmx_settings" to "fallbacks".

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,54 @@
...
+
+
+def get_theme_names():
+    themes_from_setting = set(get_themes_from_setting())
+    themes_from_css = set(get_themes_from_css())
+    installed_themes = themes_from_setting | themes_from_css
+    all_themes = themes_from_setting & themes_from_css
+    if all_themes != installed_themes:
+        raise ImproperlyConfigured("Themes in settings is out of sync with themes installed")
+    return installed_themes
```
elfjes, line 50: nitpick: I think there is a mixup between the names installed_themes and all_themes. `|` is the set union, which includes all items in either set, while `&` is the set intersection which only includes items that are in both sets. the union is a superset of the intersection. It makes more sense that `all_themes` is the superset and `installed_themes` the subset.

In this case there this has no effect on the functionality though


you could also use `^` to get the symmetric difference of both sets. If it's not empty, it's ImproperlyConfigured

```suggestion
    if themes_from_setting ^ themes_from_css:
        raise ImproperlyConfigured("Themes in settings is out of sync with themes installed")
    return themes_from_setting 
```

src/argus_htmx/checks.py
```diff
@@ -0,0 +1,30 @@
...
+def check_for_valid_themes_list(app_configs, **kwargs):
+    errors = []
+    themes = []
+    try:
+        themes = get_theme_names()
+    except ImproperlyConfigured as e:
+        errors.append(
+            Error(
+                str(e),
+                hint="Regenerate styles.css",
```
elfjes, line 17:
```suggestion
                hint=f"Regenerate {get_stylesheet_path()}",
```

src/argus_htmx/checks.py
```diff
@@ -0,0 +1,30 @@
...
+                id="argus_htmx.T001",
+            )
+        )
+    else:
+        if not themes:
+            errors.append(
+                Error(
+                    "no themes installed",
+                    hint='Check the settings "DAISYUI_THEMES" and "TAILWIND_THEME_OVERRIDE" and regenerate styles.css',
```
elfjes, line 26:
```suggestion
                    hint=f'Check the settings "DAISYUI_THEMES" and "TAILWIND_THEME_OVERRIDE" and regenerate {get_stylesheet_path()}',
```

src/argus_htmx/checks.py
```diff
@@ -0,0 +1,30 @@
...
+                    id="argus_htmx.T002",
+                )
+            )
+    return errors
```
elfjes, line 30: a simplification:

```suggestion
@register
def check_for_valid_themes_list(app_configs, **kwargs):
    try:
        themes = get_theme_names()
    except ImproperlyConfigured as e:
        return [
            Error(
                str(e),
                hint=f"Regenerate {get_stylesheet_path()}",
                id="argus_htmx.T001",
            )
        ]

    if not themes:
        return [
            Error(
                "no themes installed",
                hint=f'Check the settings "DAISYUI_THEMES" and "TAILWIND_THEME_OVERRIDE" and regenerate {get_stylesheet_path()}',
                id="argus_htmx.T002",
            )
        ]

    return []
```

elfjes (changes requested):
Overall I like this.

There's one thing though. it seems that `check_for_valid_themes_list` is scheduled to run after the Models are instantiated. which means that `argus_htmx.themes.constants.py` is imported, which runs `get_theme_names`, which throws the ImproperlyConfigured error and now we don't get our nice error message.

elfjes:
Also lolololol the check breaks the `tailwind_config` command so we can't regenerate 😂 

maybe it should be a warning instead?

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,54 @@
...
+from re import findall
+
 from django.conf import settings
-from argus_htmx import settings as argus_htmx_settings
+from django.core.exceptions import ImproperlyConfigured
 
+from argus_htmx import settings as fallbacks
 
-def get_themes():
-    return getattr(settings, "DAISYUI_THEMES", argus_htmx_settings.DAISYUI_THEMES)
```
elfjes, line 6: tailwind_config depends on this function so we need to keep it (perhaps this one should be `get_themes_from_setting` and the other one `get_theme_names_from_settings`

src/argus_htmx/checks.py
```diff
@@ -0,0 +1,27 @@
...
+            )
+        ]
+    if not themes:
+        return [
+            Error(
+                "no themes installed",
+                hint='Check the settings "DAISYUI_THEMES" and "TAILWIND_THEME_OVERRIDE" and regenerate styles.css',
+                id="argus_htmx.T002",
+            )
+        ]
```
elfjes, line 27:
```suggestion
        ]
    return []
```

src/argus_htmx/themes/utils.py
```diff
@@ -1,21 +1,63 @@
...
+LOG = logging.getLogger(__name__)
 
 
-def get_theme_names():
-    themes = get_themes()
+def get_themes_from_setting():
+    themes_setting = getattr(settings, "DAISYUI_THEMES", fallbacks.DAISYUI_THEMES)
```
elfjes, line 22: `tailwind_config` needs a function that just exposes the setting

elfjes (changes requested):
Thoughts about using `get_stylesheet_path()` in the error/warning messages?

src/argus_htmx/themes/utils.py
hmpf: Fixed in later commits

elfjes (approved):

PR #227, opened 2024-11-06
Make sure to reset checkbox state on some browsers (eg. firefox)

elfjes:
On some browsers (at least Firefox), the multiselect checkbox state is remembered after a bulk action, depending on the current incident filter, the remberbed selected checkboxes may now be for a completely different incident. We should make sure to reset the checkbox state


elfjes:
> The same should be added to `src/argus_htmx/templates/htmx/incidents/_selected_incidents_header.html` so that the select-all checkbox is also reset

fixed

PR #230, opened 2024-11-08
Systematize names: some easy pickings

hmpf:
Whether the name of something is singular or plural is a bit of a mess. Django custom is singular for models, modelname_list in template names that handles more than one of modelnames instances.

These are the easiest pickings.

Bonus: renaming the template for the checkbox to select all incidents to a name that better reflects that.

What remains are:

* themes
* incidents
* destinations

Renaming `htmx/incidents` to `htmx/incident` will be the big one.

"preferences" (not merged yet) stays plural since one instance of the Preferences model contains multiple preferences. (Maybe the Preferences-model could be renamed to "PreferenceNamespace").


hmpf:
Moved to Uninett/Argus#1015

PR #233, opened 2024-11-12
Add THEME_DEFAULT setting

elfjes:


hmpf (approved):
I hope there is a way for htmx to snitch about whether a browser is set to dark-mode or light-mode but this is a good first step.

PR #235, opened 2024-11-13
Fix regression in htmx response to theme update

elfjes:
a regression in https://github.com/Uninett/argus-htmx-frontend/pull/212/ caused the change theme view to return an incorrect resopnse  which broke subsequent theme changes 
see https://github.com/Uninett/argus-htmx-frontend/pull/212/files#diff-01ed40f6dbfb6d8070dccb23a1c34c1d819fb9175eace41c7cd175f64b0d10fcR47

### DIFF
--- a/src/argus_htmx/themes/views.py
+++ b/src/argus_htmx/themes/views.py
@@ -7,7 +7,7 @@ from django.views.decorators.http import require_GET, require_POST
 from django.http import HttpResponse, HttpResponseRedirect
 from django_htmx.http import HttpResponseClientRefresh
 
-from argus.auth.utils import get_preference, save_preference
+from argus.auth.utils import save_preference
 
 from argus_htmx.constants import THEME_NAMES
 from argus_htmx.incidents.views import HtmxHttpRequest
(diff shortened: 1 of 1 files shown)


src/argus_htmx/themes/views.py
```diff
@@ -40,9 +40,7 @@ def theme_names(request: HtmxHttpRequest) -> HttpResponse:
 
 @require_POST
 def change_theme(request: HtmxHttpRequest) -> HttpResponse:
-    theme = get_preference(request, "argus_htmx", "theme")
     success = save_preference(request, request.POST, "argus_htmx", "theme")
     if success:
-        theme = get_preference(request, "argus_htmx", "theme")
-        return HttpResponse(theme)
```
hmpf, line 47: How is this a regression? change_theme has always returned the new theme name on success, the theme dropdown does not collapse after changing.

elfjes: Because we are appending the messages to the response we need to be sure that the response is a valid partial html. I therefore wrapped the theme in a span using the `_current_theme.html ` template

hmpf: Ah, thx

PR #237, opened 2024-11-14
Bulk action loading indicator

elfjes:
Adds a loading indicator to the _base_form_modal that can be toggle using `hx-indicator` and adds this `hx-indicator` to the bulk modals


elfjes:
moved to https://github.com/Uninett/Argus/pull/1022

PR #240, opened 2024-11-15
Polish setup documentation in README

podliashanyk:


hmpf (changes requested):
The README is basically not in use anymore save for the new migration guide. Please move this to the correct locations in `argus-server/docs`.

PR #242, opened 2024-11-26
add date-only and time-only formats to preferences

elfjes:


hmpf (changes requested):
Add docs (into docs/-tree here with a name that does not clash with any file in argus-server docs/) and add a link to the python docs for datetime formats.

hmpf (approved):

